PR #1, opened 2020-07-26
Added tests, IPython pretty repr, and some func signature preservation

ryxcommar:
- Created `.gitignore` from https://github.com/github/gitignore/blob/master/Python.gitignore
- Created `setup.py` to support `pip install -e .`
- Added `_repr_latex_()` to the wrapped class. This lets it render in Jupyter notebooks without having to wrap in `Math(str(solve))`. Now simply typing a function e.g. `solve` in a Jupyter notebook cell renders the LaTeX.
- Added some function signature preservation with `.__doc__` and `.__name__` properties in the wrapped class.
- PEP8'ed `repr` -> `repr_` to avoid namespace conflict with the builtin `repr` function.
- Added unit-tests with `pytest`. Asserts the behaviors specified in https://twitter.com/odashi_t/status/1286958318884249600, plus asserts `_repr_latex_()` returns the same string. Unit tests can be run with `pip install pytest` -> `python -m pytest ./tests` or alternatively `pip install pytest` -> `pip install -e .` -> `pytest`.
- Added more trigonometry: `tan` and `cos`.
- Added `requirements.txt`. Dependencies are solely for Pytest.

I love this module, it's superb. I hope it has a good future.

### DIFF
--- a/latexify.py
+++ b/latexify.py
@@ -52,10 +52,10 @@ class LatexifyVisitor(ast.NodeVisitor):
 
   def visit_UnaryOp(self, node):
     def _wrap(child):
-      repr_ = self.visit(child)
+      latex = self.visit(child)
       if isinstance(child, ast.BinOp) and child.op in (ast.Add, ast.Sub):
-        return '(' + repr_ + ')'
-      return repr_
+        return '(' + latex + ')'
+      return latex
 
     reprs = {
         ast.UAdd: (lambda: _wrap(node.operand)),
@@ -80,13 +80,13 @@ class LatexifyVisitor(ast.NodeVisitor):
       return self.visit(child)
 
     def _wrap(child):
-      repr_ = _unwrap(child)
+      latex = _unwrap(child)
       if isinstance(child, ast.BinOp):
         cp = priority[type(child.op)] if type(child.op) in priority else 100
         pp = priority[type(node.op)] if type(node.op) in priority else 100
         if cp < pp:
-          return '(' + repr_ + ')'
-      return repr_
+          return '(' + latex + ')'
+      return latex
 
     l = node.left
     r = node.right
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -0,0 +1,13 @@
+atomicwrites==1.4.0
```
odashi, line 1: This looks to require too many dependencies that is not used in the library. If they are not imported directly it is good to omit from this.

setup.py
```diff
@@ -0,0 +1,19 @@
...
+
+setup(
+  name='Latexify',
+  url='https://github.com/odashi/latexify_py',
+  py_modules=['latexify'],
+  license='Apache License 2.0',
+  author='Yusuke Oda',
+  description='Generates LaTeX math description from Python functions.',
+  long_description=readme,
+  python_requires='>=3.6',
```
odashi, line 17: As `ast` library is frequently changing, we could not provide supports for far-newer versions. I think it is still compatible as of v3.8, `>=3.6, <=3.8` is good for now.

latexify.py
```diff
@@ -50,10 +52,10 @@ def visit_Num(self, node):
 
   def visit_UnaryOp(self, node):
     def _wrap(child):
-      repr = self.visit(child)
+      repr_ = self.visit(child)
```
odashi, line 55: I guess just `latex` is better than `repr`.

setup.py
ryxcommar: Done.

requirements.txt
ryxcommar: Done.

latexify.py
ryxcommar: Done.

ryxcommar:
Oops, specifier doesn't work, returns error. Needs to be:

```python
  python_requires='>=3.6.*, <3.9.*',
``

I am sorry about that.

Issue #6, opened 2020-07-26
Greek letters, \bar{}, and \hat{}


ryxcommar:
Something like this would be a good idea:

![image](https://user-images.githubusercontent.com/53090272/88488823-e3dc5f80-cf5d-11ea-9ab6-2b0a58dc48bf.png)

I have an implementation here, but I am not sure if it is good, or if it will do unexpected things:

https://github.com/ryxcommar/latexify_py/tree/master

odashi:
Yes indeed. This kind of modification should be optional, or it will include unintended changes, e.g.:

- `xi` ... `\xi` or `x_i`
- `\hat{y}` or `y_{hat}`

Especially the second case is quite troublesome, as the underscore (`_`) completely behaves differently.

For the interface, some optional parameter of `with_latex`/`get_latex` would help to switch the behavior:
```
@with_latex(math_symbol=True, diacritic=True)
def foobar(...):
  ...
```
I think `math_symbol` is okay to add, but consider `diacritic` is not for now. Probably we need to support subscripts before diacritics.

ryxcommar:
Something like this?

Behaviors:

![image](https://user-images.githubusercontent.com/53090272/88491975-35441900-cf75-11ea-9bcb-7b86d8693900.png)

Code:

```python
# ...
from typing import Callable

# ...

class LatexifyVisitor(ast.NodeVisitor):

  def __init__(self, math_symbol=True):
    self.math_symbol = math_symbol
    super(ast.NodeVisitor).__init__()

  def _parse_math_symbols(self, val: str) -> str:
    if not self.math_symbol:
      return val
    else:
      greek_and_hebrew = [
          'aleph', 'alpha', 'beta', 'beth', 'chi', 'daleth',
          'delta', 'digamma', 'epsilon', 'eta', 'gamma', 'gimel',
          'iota', 'kappa', 'lambda', 'mu', 'nu', 'omega', 'omega',
          'phi', 'pi', 'psi', 'rho', 'sigma', 'tau', 'theta',
          'upsilon', 'varepsilon', 'varkappa', 'varphi', 'varpi', 'varrho',
          'varsigma',
          'vartheta', 'xi', 'zeta'
      ]
      if val.lower() in greek_and_hebrew:
        return '\\' + val + ' '
      else:
        return val

  # ........

  def visit_FunctionDef(self, node):
    name_str = r'\operatorname{' + str(node.name) + '}'
[...]

PR #8, opened 2020-07-27
math_symbol kwarg

ryxcommar:
re: issue #6 

# `math_symbol` kwarg

New behaviors:

image

Old functions remain the same, so long as variable names are not in the `greek_and_hebrew` list.

image

# Fixes to `setup.py`

Moving from a module structure to a package structure, `py_modules=['latexify']` no longer works. So I replaced it with `packages=find_packages()`.

Additionally, I tested that `<=3.8` does not work as intended. (Technically, I tested that `<=3.7` does not work, as I am on python 3.7). I am using `<3.9` now instead, which should work.

### DIFF
--- a/latexify/core.py
+++ b/latexify/core.py
@@ -182,7 +182,7 @@ def get_latex(fn, math_symbol=True):
   return LatexifyVisitor(math_symbol=math_symbol).visit(ast.parse(inspect.getsource(fn)))
 
 
-def with_latex(math_symbol=True):
+def with_latex(*args, math_symbol=True):
 
   class _LatexifiedFunction:
     def __init__(self, fn):
@@ -217,6 +217,7 @@ def with_latex(math_symbol=True):
       """
       return self._str
 
-  def _wrapper(fn):
-    return _LatexifiedFunction(fn)
-  return _wrapper
+  if len(args) == 1 and callable(args[0]):
+    return _LatexifiedFunction(args[0])
+  else:
+    return lambda fn: _LatexifiedFunction(fn)
--- a/latexify/core.py
+++ b/latexify/core.py
@@ -3,12 +3,11 @@
 import ast
 import math
 import inspect
-from typing import Callable
 
 
 class LatexifyVisitor(ast.NodeVisitor):
 
-  def __init__(self, math_symbol=True):
+  def __init__(self, math_symbol):
     self.math_symbol = math_symbol
     super(ast.NodeVisitor).__init__()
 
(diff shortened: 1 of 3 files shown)


latexify/core.py
```diff
@@ -193,4 +215,7 @@ def _repr_latex_(self):
       """
       return self._str
 
-  return _LatexifiedFunction(fn)
+  if len(args) == 1 and isinstance(args[0], Callable):
```
odashi, line 218: Use `callable(args[0])` instead.

latexify/core.py
```diff
@@ -3,9 +3,31 @@
 import ast
 import math
 import inspect
+from typing import Callable
 
 
 class LatexifyVisitor(ast.NodeVisitor):
+
+  def __init__(self, math_symbol=True):
```
odashi, line 11: Remove the default value. It should be specified in only the user interfaces.

latexify/core.py
```diff
@@ -3,9 +3,31 @@
...
+      return val
+    greek_and_hebrew = [
+        'aleph', 'alpha', 'beta', 'beth', 'chi', 'daleth',
+        'delta', 'digamma', 'epsilon', 'eta', 'gamma', 'gimel',
+        'iota', 'kappa', 'lambda', 'mu', 'nu', 'omega', 'omega',
+        'phi', 'pi', 'psi', 'rho', 'sigma', 'tau', 'theta',
+        'upsilon', 'varepsilon', 'varkappa', 'varphi', 'varpi', 'varrho',
+        'varsigma', 'vartheta', 'xi', 'zeta'
+    ]
+    if val.lower() in greek_and_hebrew:
```
odashi, line 26: `lower()` looks a too generalized behavior. I think we need to process only exact matches.

latexify/core.py
```diff
@@ -3,9 +3,31 @@
...
+  def __init__(self, math_symbol=True):
+    self.math_symbol = math_symbol
+    super(ast.NodeVisitor).__init__()
+
+  def _parse_math_symbols(self, val: str) -> str:
+    if not self.math_symbol:
+      return val
+    greek_and_hebrew = [
+        'aleph', 'alpha', 'beta', 'beth', 'chi', 'daleth',
+        'delta', 'digamma', 'epsilon', 'eta', 'gamma', 'gimel',
```
odashi, line 20: `gamma` hurts `math.gamma` and it needs some workaround (just replacing `math.gamma` with `math.{\\gamma}`?).

latexify/core.py
```diff
@@ -154,16 +176,16 @@ def visit_If(self, node):
...
 
 
-def get_latex(fn):
-  return LatexifyVisitor().visit(ast.parse(inspect.getsource(fn)))
+def get_latex(fn, math_symbol=True):
+  return LatexifyVisitor(math_symbol=math_symbol).visit(ast.parse(inspect.getsource(fn)))
 
 
-def with_latex(fn):
+def with_latex(*args, math_symbol=True):
```
odashi, line 183: This does not work as a decorator. Fix this as:

```
def with_latex(math_symbol=True):
  class _LatexifiedFunction:
    ...
  def _wrapper(fn):
    return _LatexifiedFunction(fn)
  return _wrapper
```

ryxcommar: Your suggestion will change the behavior so that this happens, which is incompatible with the original API:

image

The original I had supports both `@with_latex(**stuff)` and `@with_latex`:

image

The major downside of the approach I used is that the decorator's signature looks ugly, and in some very rare cases (usually when the user is doing something weird and not advisable) it can break.

I believe the authors of [this module](https://github.com/micheles/decorator) came up with a good solution to this problem, but I am trying to not put in any external dependencies.

ryxcommar: I have implemented the change as per your request, I just want to be clear about what this change will do to the original API.

odashi: Ah I see, I missed the last case and your impl. is actually better. There's no ploblem to revive the original one.

ryxcommar: Reimplemented!

ryxcommar:
Sorry about that. Fixed, and double checked that tests pass.

odashi:
Great work, thanks!

PR #9, opened 2020-07-27
fixed setup.py

ryxcommar:
I'm so sorry! 😱 😱 😱 😱 I broke `setup.py`! I was wrong: `py_modules=['latexify']` still works!

I did not notice this because the error does not come up unless you work in a totally new environment.

(When I tested my `setup.py` code, the `import latexify` just used the old install, and did not return `ModuleNotFoundError`.)

image

However it works if I do this:

image

I'm super sorry about this. 😱

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,5 @@
 import io
 from setuptools import setup
-from setuptools import find_packages
 
 
 with io.open('README.md', 'rt', encoding='utf8') as f:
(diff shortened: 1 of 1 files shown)


odashi:
Ah okay. This is also my fault.
I got some suspicion since `foo.py` and `foo/__init__.py` works similarly, but passed the change without any inquiry.

odashi:
Related to #8.

PR #15, opened 2020-08-01
Add More Functions

1MLightyears:
Add support for <,>,>=,<=,!=,is; Add support for complex boolean statements; leave interface for changing "\triangleq"

I ran test_io.py and no assert is triggered. But since new functions are added, maybe I should also add test cases there?

### DIFF
--- a/latexify/core.py
+++ b/latexify/core.py
@@ -175,11 +175,8 @@ class LatexifyVisitor(ast.NodeVisitor):
       return lstr + r'\le ' + rstr
     elif isinstance(node.ops[0], ast.NotEq):
       return lstr + r'\ne ' + rstr
-    #TODO: Needs further discussion for is,isnot & \in,\notin
     elif isinstance(node.ops[0], ast.Is):
-      return lstr + r'\space is\space ' + rstr
-    elif isinstance(node.ops[0], ast.IsNot):
-      return lstr + r'\space is\space not\space ' + rstr
+      return lstr + r'\equiv' + rstr
 
     else:
       return r'\operatorname{unknown\_comparator}(' + lstr + ', ' + rstr + ')'
--- a/latexify/core.py
+++ b/latexify/core.py
@@ -36,11 +36,15 @@ class LatexifyVisitor(ast.NodeVisitor):
   def visit_Module(self, node):
     return self.visit(node.body[0])
 
-  def visit_FunctionDef(self, node):
+  def visit_FunctionDef(self, node, equal:str=r'\triangleq'):
     name_str = r'\operatorname{' + str(node.name) + '}'
     arg_strs = [self._parse_math_symbols(str(arg.arg)) for arg in node.args.args]
     body_str = self.visit(node.body[0])
-    return name_str + '(' + ', '.join(arg_strs) + r') \triangleq ' + body_str
+    # CHANGE:(20200801@1MLightyears)
+    # use r"\triangleq" as default is fine but maybe it is more user-friendly
+    # to leave an interface for users to choose what equal mark they want to.
+    # I leave it here as-is and may implement this in future.
+    return name_str + '(' + ', '.join(arg_strs) + r') ' + equal + ' ' + body_str
 
   def visit_Return(self, node):
     return self.visit(node.value)
@@ -81,7 +85,7 @@ class LatexifyVisitor(ast.NodeVisitor):
     else:
       if callee_str.startswith('math.'):
         callee_str = callee_str[5:]
-      lstr = r'\operatorname{' + callee_str + '}\left('
+      lstr = r'\operatorname{' + callee_str + r'}\left('
       rstr = r'\right)'
 
     arg_strs = [self.visit(arg) for arg in node.args]
@@ -163,8 +167,43 @@ class LatexifyVisitor(ast.NodeVisitor):
 
     if isinstance(node.ops[0], ast.Eq):
       return lstr + '=' + rstr
+    # CHANGE:(20200801@1MLightyears) Add support for conditions for <,>,\ge,\le,\ne,is
+    elif isinstance(node.ops[0], ast.Gt):
+      return lstr + '>' + rstr
+    elif isinstance(node.ops[0], ast.Lt):
+      return lstr + '<' + rstr
+    elif isinstance(node.ops[0], ast.GtE):
+      return lstr + r'\ge ' + rstr
+    elif isinstance(node.ops[0], ast.LtE):
+      return lstr + r'\le ' + rstr
+    elif isinstance(node.ops[0], ast.NotEq):
+      return lstr + r'\ne ' + rstr
+    elif isinstance(node.ops[0], ast.Is):
+      return lstr + r'\space is\space ' + rstr
+    elif isinstance(node.ops[0], ast.IsNot):
+      return lstr + r'\space is\space not\space ' + rstr
+
     else:
       return r'\operatorname{unknown\_comparator}(' + lstr + ', ' + rstr + ')'
+
+  # CHANGE:(20200801@1MLightyears) Add support for BoolOp, which allow Latexify to
+  # deal with more complex boolean statements.
+  # The structure is weak, for I seldom use very complex boolean statements and
+  # currently have no idea how to simply implement a result like "x=0,1" instead of
+  # "x=0 Or x=1" as it is now XD
+  def visit_BoolOp(self, node):
+    '''
+    When node is an instance of ast.BoolOp, it has:
+    node.op(ast.And, ast.Or, ast.Not): type of the boolean operator.
+    node.values(list of ast.Compare, ast.BoolOp): a list of boolean sub-clause.
+    '''
+    logic_operator = r' \operatorname{unknown\_comparator} '
+    if isinstance(node.op, ast.Or):
+      logic_operator = r'\\&\mathrm{Or}\space '
+    elif isinstance(node.op, ast.And):
+      logic_operator = r'\\&\mathrm{And}\space '
+    # visit all the elements in the ast.If node recursively
+    return logic_operator.join([self.visit(subnode) for subnode in node.values])
 
   def visit_If(self, node):
     latex = r'\left\{ \begin{array}{ll} '
--- a/latexify/core.py
+++ b/latexify/core.py
@@ -37,15 +37,11 @@ class LatexifyVisitor(ast.NodeVisitor):
   def visit_Module(self, node):
     return self.visit(node.body[0])
 
-  def visit_FunctionDef(self, node, equal:str=r'\triangleq'):
+  def visit_FunctionDef(self, node):
     name_str = r'\operatorname{' + str(node.name) + '}'
     arg_strs = [self._parse_math_symbols(str(arg.arg)) for arg in node.args.args]
     body_str = self.visit(node.body[0])
-    # CHANGE:(20200801@1MLightyears)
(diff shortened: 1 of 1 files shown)


1MLightyears:
oops a minor mistake
in core.py, at line 200:
```
    logic_operator = r' \operatorname{unknown\_comparator} '
```
should be 
```
    logic_operator = r' \operatorname{unknown\_operator} '
```
I copied from above and forgot to modify it.
sorry...

latexify/core.py
```diff
@@ -163,9 +167,44 @@ def visit_Compare(self, node):
...
+  # "x=0 Or x=1" as it is now XD
+  def visit_BoolOp(self, node):
+    '''
+    When node is an instance of ast.BoolOp, it has:
+    node.op(ast.And, ast.Or, ast.Not): type of the boolean operator.
+    node.values(list of ast.Compare, ast.BoolOp): a list of boolean sub-clause.
+    '''
+    logic_operator = r' \operatorname{unknown\_comparator} '
+    if isinstance(node.op, ast.Or):
+      logic_operator = r'\\&\mathrm{Or}\space '
```
odashi, line 202: LaTeX has logical binary operators `\lor` and `\land`.

latexify/core.py
```diff
@@ -163,9 +167,44 @@ def visit_Compare(self, node):
...
+
     else:
       return r'\operatorname{unknown\_comparator}(' + lstr + ', ' + rstr + ')'
 
+  # CHANGE:(20200801@1MLightyears) Add support for BoolOp, which allow Latexify to
+  # deal with more complex boolean statements.
+  # The structure is weak, for I seldom use very complex boolean statements and
+  # currently have no idea how to simply implement a result like "x=0,1" instead of
+  # "x=0 Or x=1" as it is now XD
+  def visit_BoolOp(self, node):
```
odashi, line 194: This function needs the same strategy to `visit_BinOp` to deal with parenthesis appropriately (generally, logical operators has priorities: *not* > *and* > *or*)

latexify/core.py
```diff
@@ -163,9 +167,44 @@ def visit_Compare(self, node):
...
+    '''
+    When node is an instance of ast.BoolOp, it has:
+    node.op(ast.And, ast.Or, ast.Not): type of the boolean operator.
+    node.values(list of ast.Compare, ast.BoolOp): a list of boolean sub-clause.
+    '''
+    logic_operator = r' \operatorname{unknown\_comparator} '
```
odashi, line 200: Put this line on the else block.

latexify/core.py
```diff
@@ -36,11 +36,15 @@ def generic_visit(self, node):
   def visit_Module(self, node):
     return self.visit(node.body[0])
 
-  def visit_FunctionDef(self, node):
+  def visit_FunctionDef(self, node, equal:str=r'\triangleq'):
     name_str = r'\operatorname{' + str(node.name) + '}'
     arg_strs = [self._parse_math_symbols(str(arg.arg)) for arg in node.args.args]
     body_str = self.visit(node.body[0])
-    return name_str + '(' + ', '.join(arg_strs) + r') \triangleq ' + body_str
+    # CHANGE:(20200801@1MLightyears)
```
odashi, line 43: Remove all `CHANGE:` comments to keep the code simple. This kind of comments are unnecessary since commit messages should provide enough information.

latexify/core.py
```diff
@@ -163,9 +167,44 @@ def visit_Compare(self, node):
...
       return r'\operatorname{unknown\_comparator}(' + lstr + ', ' + rstr + ')'
 
+  # CHANGE:(20200801@1MLightyears) Add support for BoolOp, which allow Latexify to
+  # deal with more complex boolean statements.
+  # The structure is weak, for I seldom use very complex boolean statements and
+  # currently have no idea how to simply implement a result like "x=0,1" instead of
+  # "x=0 Or x=1" as it is now XD
+  def visit_BoolOp(self, node):
+    '''
+    When node is an instance of ast.BoolOp, it has:
```
odashi, line 196: This information is already provided by `ast`. We need not to provide the same stuff.

odashi:
Yes, tests should be updated, but the test directory is not well organized for now.

latexify/core.py
```diff
@@ -36,11 +36,15 @@ def generic_visit(self, node):
   def visit_Module(self, node):
     return self.visit(node.body[0])
 
-  def visit_FunctionDef(self, node):
+  def visit_FunctionDef(self, node, equal:str=r'\triangleq'):
```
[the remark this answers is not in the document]
1MLightyears: for "equal": 
the fact is that I haven't decided how to implement this meanwhile keep the current interface work (maybe allowing @with_latex(equal="\equal") or something like that will be a good idea). I add this only because I seldom use "\triangleq", and I feel it necessary to have "\equal" for some occasions.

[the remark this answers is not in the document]
1MLightyears: I will try to make this work in the next commit.

latexify/core.py
```diff
@@ -163,9 +167,44 @@ def visit_Compare(self, node):
...
+      return lstr + '>' + rstr
+    elif isinstance(node.ops[0], ast.Lt):
+      return lstr + '<' + rstr
+    elif isinstance(node.ops[0], ast.GtE):
+      return lstr + r'\ge ' + rstr
+    elif isinstance(node.ops[0], ast.LtE):
+      return lstr + r'\le ' + rstr
+    elif isinstance(node.ops[0], ast.NotEq):
+      return lstr + r'\ne ' + rstr
+    elif isinstance(node.ops[0], ast.Is):
```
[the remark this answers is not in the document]
1MLightyears: for "is"&"is not":
I tested several cases-the layout, the view, the font, of "is" and "is not", and found that what I really need may actually be "\in" and "\notin". Because normal papers would almost never have statements like "$if x is int$", but "$if x \in Z$". 
An approach is that the user manually define a variable like `Z=type('Z',(int,),{})` and then use ` if x is Z` to have the output "$if x is Z$". So maybe change "$is$" and "$is not$" to "$\in$" and "$\notin$" is better. 
I cannot decide this alone, so it's great to make some discussion for this.

latexify/core.py
1MLightyears: OMG I even never heard them!
Thank you for teaching me these. I will change them at the next commit.

latexify/core.py
1MLightyears: for "visit_BoolOp"
At the first version of `visit_BoolOp`, I actually wrote the `not` parts. But after testing a few cases I found it kind of silly to have output like "$if not x\equal 10$", it's not something would appear in anyone's paper. Instead what we actually want is "$if x\neq 10$", but that needs a new list or dict to detect and find proper negative statements. I will work on that after next commit.
About the parenthesis, it is true that parenthesis needs to be mentioned, but when it comes to LaTeX output it needs to be reader-friendly. One approach is to add brackets, another is to start a new line when indicating priority like:
```
if x < 1 \lor x > 10\\
\land x>0
```
for `if (x<1 or x>10)and(x>0)`. 
"$\lor$" and "$\land$" seem not so good at such case btw.

1MLightyears: As far as the papers I read, most authors tend to use comma to connect statements have AND relationship, because OR relationship is quite rarely seen. ORs are likely to be clearly indicated when needed.
If you have a better output format for this plz reply me, and I will try to implement the brackets for parenthesis in next commit.

latexify/core.py
1MLightyears: OK. Change that in the next commit.

latexify/core.py
1MLightyears: for "CHANGE"
ah... sorry to bring in my personal habit. I used to work without git.
Will be removed in the next commit.

latexify/core.py
1MLightyears: Forget to delete one of the ''' comments - will be delete in the next commit, sorry!

ryxcommar:
Yes I agree, tests aren't very well conceptualized and organized right now. I just threw them in because it's good to have some tests and "some tests" > "no tests". But I did not spend a lot of time making sure it was extremely scalable.

I personally would recommend covering new functionality under the current setup as it's still really small and easy to manage, but eventually that file is going to get too disorganized for its own good. So someone with more experience writing tests should definitely come up with a better system than the one I implemented.

As a note, I also used pytest because I know Google uses it a lot in their GCP products; so I figured @odashi would be familiar with it and prefer it. But I never consulted him about that.

latexify/core.py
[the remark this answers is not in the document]
odashi: I meant this option could *never* be used by users: all `visit_***` functions are called internally by `NodeVisitor` with only `node` parameter. We have basically no way to pass extra arguments.

latexify/core.py
odashi: I prefer single-line form with `\neg`, `\lor` and `\land`. It keeps the same syntax to Python except appearances of operators, and would eventually not surprise users. I think we have to keep minimality about extra conventions of LaTeX since any styles are basically not a common consensus of us (though there are already some strong assumptions introduced when I wrote this first).

odashi:
@1MLightyears You can push another commit(s) to this request to resolve comments.

odashi:
@ryxcommar Any test frameworks are okay. I personally prefer `unittest` if there's no particular intent, as it is a part of Python distribution.

latexify/core.py
1MLightyears: I agree with the keeping minimality about extra conventions. I will try to implement the brackets for parenthesis in single-line form in the next commit.

latexify/core.py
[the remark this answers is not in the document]
1MLightyears: yes I made this mistake because I've just started learning ast and am not so familiar with its logic. Will remove it.
But I still feel it necessary to leave interface for users to change the form of equal mark. If latexify aims to work on the most cases, and "keep minimality about extra conventions"-I strongly agree with this-should be made into practice, then latexify should leave the most freedom for users to customize their output.
If you also agree with this, I will try to find my approach to implement this with no original interface changed.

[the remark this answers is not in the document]
odashi: I have suggested to add a similar option like `math_symbol`. It should work and not affect the behavior of existing usages.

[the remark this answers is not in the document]
1MLightyears: OK. I take that as the final result of discussion and will try to implement it in the next commit.

latexify/core.py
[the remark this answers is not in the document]
odashi: How is this going?

For the use-cases like `x is int`, the equivalent Python functionality is `isinstance(x, int)`. The Python's `is` is not a `belongs-to (\in)` operator: it checks if the two expressions represent the same instance, e.g., `1 is 1` is `True`. Also the Python's `in` is provided basically for collection types. If we want to implement the `\in` operator, I think we need to take two implementations: the Python's `in` and the pattern match replacement for `isinstance`. As the latter one may involve some garbages, it may be good to restrict acdeptable types (e.g., number types only).

If we provided some representation of `is`, I think the nearest one is maybe *isomorphic* (`\equiv`) though there's some differences between them.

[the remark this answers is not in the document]
1MLightyears: Oh I thought you mean not to change them at present.
OK, I will change the implementation of the `ast.Is` clause into `\equiv ` and delete the `ast.IsNot` clause. As for `isinstance`, that couldn't be processed in `visit_Compare` so maybe that might be implemented in the future.

PR #24, opened 2020-08-26
Enable assign variables

yixingfu:
Enable for example following case:

```
@latexify.with_latex
def foo(a):
 c = foo(a-1)
 return c
```

Also made changed test case a bit so that it is passing.

### DIFF
--- a/latexify/core.py
+++ b/latexify/core.py
@@ -243,7 +243,7 @@ def get_latex(fn, math_symbol=True):
     source = inspect.getsource(fn)
   except Exception:
     # Maybe running on console.
-    source = dill.source.getsource(fn, enclosing=True)
+    source = dill.source.getsource(fn)
   return LatexifyVisitor(math_symbol=math_symbol).visit(ast.parse(source))
 
 
--- a/latexify/core.py
+++ b/latexify/core.py
@@ -35,6 +35,7 @@ class LatexifyVisitor(ast.NodeVisitor):
 
   def __init__(self, math_symbol):
     self.math_symbol = math_symbol
+    self.assigned_values = {}
     super(ast.NodeVisitor).__init__()
 
   def _parse_math_symbols(self, val: str) -> str:
@@ -54,8 +55,10 @@ class LatexifyVisitor(ast.NodeVisitor):
   def visit_FunctionDef(self, node):
     name_str = r'\mathrm{' + str(node.name) + '}'
     arg_strs = [self._parse_math_symbols(str(arg.arg)) for arg in node.args.args]
-    body_str = self.visit(node.body[0])
-    return name_str + '(' + ', '.join(arg_strs) + r')\triangleq ' + body_str
+    for body_item in node.body[:-1]:
+      self.visit(body_item)
+    return_str = self.visit(node.body[-1])
+    return name_str + '(' + ', '.join(arg_strs) + r')\triangleq ' + return_str
 
   def visit_Return(self, node):
     return self.visit(node.value)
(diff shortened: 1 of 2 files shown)


yixingfu:
For the above example, is it more ideal to assign the expressions, so that we get in this case
`foo(a) \triangleq foo(a-1)`, or do something like `foo(a) \triangleq c \n where c \triangleq foo(a-1)`?

odashi:
Thanks for proposing this change. Could you please put more detailed explanation about how this change works (e.g, a generalized example(s))?

I think tweaking multiple statements is not trivial and it may involve some optimization/reduction techniques after parsing the source. It is also not trivial to decide if expanding underlying formula in the symbol e.g.,

```python
def foo(x):
  a = ...some long formula...
  return math.sin(a) * math.sin(a) + math.cos(a) * math.cos(a)
  # may generate a very long equation, but this is essentially equal to 1.
```

Though involving such techniques is interesting: it may be worthy for users to find a better math form by executing `with_latex`.

yixingfu:
> Thanks for proposing this change. Could you please put more detailed explanation about how this change works (e.g, a generalized example(s))?
> 
> I think tweaking multiple statements is not trivial and it may involve some optimization/reduction techniques after parsing the source. It is also not trivial to decide if expanding underlying formula in the symbol e.g.,
> 
> ```python
> def foo(x):
>   a = ...some long formula...
>   return math.sin(a) * math.sin(a) + math.cos(a) * math.cos(a)
>   # may generate a very long equation, but this is essentially equal to 1.
> ```
> 
> Though involving such techniques is interesting: it may be worthy for users to find a better math form by executing `with_latex`.

I totally agree that parsing multiple statements is not trivial; but we can certainly allow some simple situations while give clear error messages otherwise clarifying the limitation. 
Here the idea is to allow one (and only one) assignment of each variable. If a variable is assigned more than once, error message is produced. When the variable is assigned, we save the latex form of the variable in a dictionary; and every time the variable is used (as `visit_Name`),  it first tries to find existing expression in the dict. 

Indeed this is going to create a fully expanded formula. There could be an option to enable expansion or not. If choose to not expanding, we can use the name when `visit_Name`, and at the end append a line. For example, 
```
def foo(x):
  a = ...some long formula...
  return math.sin(a) * math.sin(a) + math.cos(a) * math.cos(a)
```
would give 
`\sin(some formula)\sin(some formula) + \cos(some formula)\cos(some formula)` if choose to expand, or 
`\sin(a)\sin(a) + \cos(a)\cos(a), where a=some formula` is choose to not expand.  What do you think?

latexify/core.py
```diff
@@ -54,8 +55,10 @@ def visit_Module(self, node):
   def visit_FunctionDef(self, node):
     name_str = r'\mathrm{' + str(node.name) + '}'
     arg_strs = [self._parse_math_symbols(str(arg.arg)) for arg in node.args.args]
-    body_str = self.visit(node.body[0])
-    return name_str + '(' + ', '.join(arg_strs) + r')\triangleq ' + body_str
+    for body_item in node.body[:-1]:
+      self.visit(body_item)
```
odashi, line 59: Directly returning math forms would be better to maintain this class being stateless.

latexify/core.py
```diff
@@ -54,8 +55,10 @@ def visit_Module(self, node):
   def visit_FunctionDef(self, node):
     name_str = r'\mathrm{' + str(node.name) + '}'
     arg_strs = [self._parse_math_symbols(str(arg.arg)) for arg in node.args.args]
-    body_str = self.visit(node.body[0])
-    return name_str + '(' + ', '.join(arg_strs) + r')\triangleq ' + body_str
+    for body_item in node.body[:-1]:
```
odashi, line 58: We need more strict check: this visitor should only allow a series of assignments and denies other statements (e.g., if-blocks, which confuses which variable is assigned, and its content).

odashi (changes requested):
The idea looks basically good. Thanks for explaining.

odashi:
what's going on?

odashi:
Close this request due to no action.

PR #27, opened 2020-09-07
Setup CI and fix tests

ShigekiKarita:
Hi @odashi 

I added CI by Github Actions https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions. Then, I fixed codes to pass `pytest` without errors and warnings. For recent deprecation in the ast module, see also https://docs.python.org/3/library/ast.html#ast.NodeVisitor
> Deprecated since version 3.8: Methods visit_Num(), visit_Str(), visit_Bytes(), visit_NameConstant() and visit_Ellipsis() are deprecated now and will not be called in future Python versions. Add the visit_Constant() method to handle all constant nodes.

### DIFF
--- a/latexify/core.py
+++ b/latexify/core.py
@@ -52,10 +52,10 @@ class LatexifyVisitor(ast.NodeVisitor):
     return self.visit(node.body[0])
 
   def visit_FunctionDef(self, node):
-    name_str = r'\mathrm{' + str(node.name) + '}'
+    name_str = r'\operatorname{' + str(node.name) + '}'
     arg_strs = [self._parse_math_symbols(str(arg.arg)) for arg in node.args.args]
     body_str = self.visit(node.body[0])
-    return name_str + '(' + ', '.join(arg_strs) + r')\triangleq ' + body_str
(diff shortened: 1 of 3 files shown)


ShigekiKarita:
PS: you can check the CI results in my forked repo https://github.com/ShigekiKarita/latexify_py/runs/1080095161?check_suite_focus=true

latexify/core.py
```diff
@@ -52,10 +52,10 @@ def visit_Module(self, node):
     return self.visit(node.body[0])
 
   def visit_FunctionDef(self, node):
-    name_str = r'\mathrm{' + str(node.name) + '}'
+    name_str = r'\operatorname{' + str(node.name) + '}'
```
odashi, line 55: Revert this: operatorname is not supported by some printing systems.

odashi:
> `visit_Constant`

Yes, this was because README said Python3.6 AST.
One concern of this library is that this is tightly depending on `ast` which is updated sometimes...

latexify/core.py
ShigekiKarita: Done

odashi (approved):

PR #29, opened 2020-09-08
Extract long table-like variables to another file

ryosuketc:
Hi @odashi 

I extracted three variables to another file for better readability.
-   _MATH_SYMBOLS (global)
-   builtin_callees (`visit_Call`)
 -   priority (`visit_BinOp`)

This is a simple refactoring with no change in its functionality and  I confirmed that it passed all tests (`pytest`).


latexify/predefined_variables.py
```diff
@@ -0,0 +1,72 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import ast
+
+def get_math_symbols():
+  math_symbols = {
```
odashi, line 18: Placing this and following dictionaries onto the module looks better rather than wrapping getter functions. Module level constants will be initialized only once and can be accessed directly from others.

latexify/predefined_variables.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2020 Google LLC
```
odashi, line 1: Simply `constants.py`?

latexify/core.py
```diff
@@ -20,16 +20,9 @@
...
-    'iota', 'kappa', 'lambda', 'mu', 'nu', 'omega', 'omega',
-    'phi', 'pi', 'psi', 'rho', 'sigma', 'tau', 'theta',
-    'upsilon', 'varepsilon', 'varkappa', 'varphi', 'varpi', 'varrho',
-    'varsigma', 'vartheta', 'xi', 'zeta', 'Delta', 'Gamma',
-    'Lambda', 'Omega', 'Phi', 'Pi', 'Sigma', 'Theta',
-    'Upsilon', 'Xi',
-}
+from latexify import predefined_variables
+
+_MATH_SYMBOLS = predefined_variables.get_math_symbols()
```
odashi, line 25: It is better to remove `_MATH_SYMBOLS` and use the imported dictionary directly.

latexify/predefined_variables.py
ryosuketc: Thank you for the feedback.
I would like to get some clarification - I should initialize the constants at the global scope of `core.py` instead of calling the getter functions at function-level to avoid being initialized multiple times? Do I understand your intention correctly?

e.g. at `core.py
```
MATH_SYMBOLS = constants.get_math_simbols()

def visit_xxx():
    some code
```

ShigekiKarita: I prefer module-level constants
```python
# constants.py
MATH_SYMBOLS = {
  ...
}
```
encouraged at https://github.com/google/styleguide/blob/gh-pages/pyguide.md#254-decision

ryosuketc: @ShigekiKarita Thank you! I guess that is what @odashi actually meant. I will go in that approach.

ryosuketc: Fixed in the following commits.

latexify/predefined_variables.py
ryosuketc: Fixed in the following commits.

latexify/core.py
ryosuketc: Fixed in the following commits.

latexify/constants.py
```diff
@@ -0,0 +1,66 @@
...
+  'math.prod': (r'\prod \left({', r'}\right)'),
+  'math.sin': (r'\sin{\left({', r'}\right)}'),
+  'math.sinh': (r'\sinh{\left({', r'}\right)}'),
+  'math.sqrt': (r'\sqrt{', '}'),
+  'math.tan': (r'\tan{\left({', r'}\right)}'),
+  'math.tanh': (r'\tanh{\left({', r'}\right)}'),
+  'sum': (r'\sum \left({', r'}\right)'),
+}
+
+BIN_OP_PRIORITY = {
```
ryosuketc, line 57: @odashi @ShigekiKarita 
`BIN_OP_PRIORITY` or `BinOp_PRIORITY`?
Should I respect style guide for constants (all capital letters) or the class name (`BinOp`) here?
https://github.com/google/styleguide/blob/gh-pages/pyguide.md#254-decision

latexify/core.py
```diff
@@ -70,34 +61,7 @@ def visit_Set(self, node):
...
-        'math.log2': (r'\log_{2}{\left({', r'}\right)}'),
-        'math.prod': (r'\prod \left({', r'}\right)'),
-        'math.sin': (r'\sin{\left({', r'}\right)}'),
-        'math.sinh': (r'\sinh{\left({', r'}\right)}'),
-        'math.sqrt': (r'\sqrt{', '}'),
-        'math.tan': (r'\tan{\left({', r'}\right)}'),
-        'math.tanh': (r'\tanh{\left({', r'}\right)}'),
-        'sum': (r'\sum \left({', r'}\right)'),
-    }
+    builtin_callees = constants.BUILTIN_CALLEES
```
ryosuketc, line 64: @odashi @ShigekiKarita 
Should I declare a local variable here to access it more simply in the function definition (line 67 - 68), or not (i.e. keep using `constants.BUILTIN_CALLEES` throughout the definition)?

In my understanding functionality remains the same.

latexify/core.py
```diff
@@ -146,16 +110,7 @@ def _wrap(child):
...
-        ast.Add: 10,
-        ast.Sub: 10,
-        ast.Mult: 20,
-        ast.MatMult: 20,
-        ast.Div: 20,
-        ast.FloorDiv: 20,
-        ast.Mod: 20,
-        ast.Pow: 30,
-    }
+    priority = constants.BIN_OP_PRIORITY
```
ryosuketc, line 113: @odashi @ShigekiKarita
Same as `constants.BUILTIN_CALLEES` - Should I declare a local variable here to access it more simply in the function definition or not?

ryosuketc (commented):
Added some comments where I would like to get feedback.

latexify/constants.py
odashi: Constants.

latexify/core.py
odashi: Usually I define aliases if something will be used many times. In this case it's basically up to you, but I think using `dict.get` seems more reasonable to avoid alias:

```python
lstr, rstr = constants.BUILTIN_CALLEES.get(callee_str, (None, None))
if lstr is None:
  ...
```

latexify/core.py
odashi: Personally it looks reasonable to introduce shorter alias.

odashi (approved):

latexify/constants.py
ryosuketc: Thanks. Then I will leave it as is.

latexify/core.py
ryosuketc: I made the suggested change.

latexify/core.py
ryosuketc: I keep it as it is for now. Please let me know in case you expected an alias shorter than the current one - like `p` or `prio`.

odashi (approved):

latexify/core.py
odashi: I meant "priority" is the shorter alias. It looks good.

odashi:
Linter warns something in `constant.py`. Please follow the guidance the CI generated.

latexify/core.py
ryosuketc: Thanks for the clarification!

ryosuketc:
@odashi 
-   I have fixed Linter errors.
-   My fork was behind the upstream. The conflicts were resolved and merged.

odashi (approved):

Issue #31, opened 2020-10-08
Underscores in function names conflict with LaTeX and shown as subscripts


ryosuketc:
Underscores in function names conflict with LaTeX and thus it is represented as subscript in LaTeX.

![image](https://user-images.githubusercontent.com/43229670/95415851-54472a00-096c-11eb-8e86-c519f61d90a9.png)

In Jupyter Notebook:
```python
@latexify.with_latex
def func_with_underscore(x):
  return 1

func_with_underscore
```

odashi:
Yeah the underscores in identifiers should be escaped normally, but in some few cases they are worthy for special treatment (as #6 discussed).

1MLightyears:
It looks like only "_" the underline is the only character that has the problem
See if I could fix it 🤔

yasirroni:
For those that come here and confused to implement it:

```
@latexify.with_latex(use_raw_function_name=True)
def forward_diff(f, h, x):
    return f(x + h) / h
```

See https://github.com/google/latexify_py/issues/82

Issue #30, opened 2020-09-30
Limits in Summations


shivam-kotwalia:
Hello Team - thanks for this wonderful package, I was wondering how do we put a limit on summation. 
![image](https://user-images.githubusercontent.com/14539580/94681234-b5ee0f80-0340-11eb-9ec7-166d8b4eb6d2.png)

Thanks in advance :)

odashi:
It is not supported for now. I guess some special treatment is needed for analyzing `sum()`.

E.g., the following expression:

```python
sum((y_bar[i] - y[i])**2 for i in range(n))
```

would represent your case, though it needs a deeper traverse over AST.

Yes it is doable, please wait a while to discuss it internally (or your PR is also welcome!)

PR #32, opened 2020-10-08
Implement summation with limits using comprehension #30

ryosuketc:
This supports summation with limits (super/subscripts) written with comprehension with for loop and range function.
Example usage is outlined in `tests/test_io.py`
 Note:
-   Indices mentioned in the issue #30 is not supported yet (e.g. `y_bar[i]`).
-   `sum_with_limit` and `test_with_latex_to_str`  in `tests/test_io.py` are impacted by issue #31 , which need to be corrected after its fix.


latexify/core.py
```diff
@@ -75,8 +75,38 @@ def visit_Call(self, node):  # pylint: disable=invalid-name
...
+      limit_str, formula_str = self.visit(node.args[0])
+      lstr =  r'\sum' + limit_str + r' \left({'
+      return lstr + formula_str + rstr
+    if callee_str == 'range':
+      return arg_strs
+
     return lstr + ', '.join(arg_strs) + rstr
 
+  def visit_GeneratorExp(self, node):  # pylint: disable=invalid-name
+    limit_str = self.visit(node.generators[0])
```
odashi, line 89: These variable names seems to be specialized for limit-like operators. General naming is better. (ditto for others)

latexify/core.py
```diff
@@ -75,8 +75,38 @@ def visit_Call(self, node):  # pylint: disable=invalid-name
...
+      if isinstance(limits, list):
+        if len(limits) == 1:
+          lower_limit, upper_limit = '0', limits[0]
+        elif len(limits) == 2:
+          lower_limit, upper_limit = limits
+      else:
+        raise ValueError
+    except ValueError as e:
+      print(e, 'Maybe function other than "range" is used is comprehension')
+    return fr'_{{{var}={lower_limit}}}^{{{upper_limit}}}'
```
odashi, line 107: This conversion should be placed at converting "sum". Comprehension itself does not know how to deal with the generated source by them.

odashi (changes requested):

ryosuketc:
This feature is pending and will be addressed after thisbchange.
As discussed offline with @odashi , we plan to revamp the implementation of `NodeVisitor` to enrich information passed between nodes.

latexify/core.py
```diff
@@ -224,6 +244,27 @@ def visit_If(self, node, action):  # pylint: disable=invalid-name
     latex += self.visit(node)
     return latex + r', & \mathrm{otherwise} \end{array} \right.'
 
+  def visit_GeneratorExp_sum(self, node):  # pylint: disable=invalid-name
```
ryosuketc, line 247: This could have been implemented with generic `visit_GeneratorExp`, which might be useful for FRs in the future (ditto for `visit_comprehension'. In that case, we might need some checks in `decorated_lstr_and_arg`.

You could view the generic implementation in this commit (sorry it's a bit dirty before refactoring).
https://github.com/google/latexify_py/pull/32/commits/0c5d67fa8aa44064b74c711a7d567d11953877ec

latexify/core.py
```diff
@@ -82,6 +82,26 @@ def visit_Set(self, node, action):  # pylint: disable=invalid-name
 
   def visit_Call(self, node, action):  # pylint: disable=invalid-name
     """Visit a call node."""
+
+    def _decorated_lstr_and_arg(node, callee_str, lstr):
+      """Decorate lstr and get its associated arguments"""
+      if callee_str == 'sum' and isinstance(node.args[0], ast.GeneratorExp):
```
ryosuketc, line 88: The more functions we support decorations for, the bigger `visit_Call' will be. Is acceptable or is there any better way to implement this type of decorations keeping future expandability in mind?

latexify/core.py
odashi: What are FRs?

This function can be reused with other operators (e.g., products and integrals have the same syntax). We can use some general action name rather than `sum`.

latexify/core.py
odashi: We can separate some routines out of this method if needed. But it looks reasonable to keep the current implementation because we are supporting only `sum`; it is too early to generalize the behavior.

latexify/core.py
odashi: I think the bare visitor (=visitor with `None` action) of every node should return only a LaTeX string. Otherwise every parent node is required to know what kind of objects are returned from children and how to process them to construct LaTeX strings.

tests/regression_test.py
```diff
@@ -50,11 +50,32 @@ def xtimesbeta(x, beta):
...
+)
+
+
+def sum_with_limit_two_args(a, n):
+  return sum(i**2 for i in range(a, n))
+
+
+sum_with_limit_two_args_latex = (
+  r'\mathrm{sum_with_limit_two_args}(a, n) '
+  r'\triangleq \sum_{i=a}^{n} \left({i^{2}}\right)'
```
odashi, line 68: `n` must be `n - 1`.

latexify/core.py
ryosuketc: > What are FRs?
Sorry for not being clear, but I meant feature requests. I thought some operations could use `GeneratorExp` and there might not be a reason to limit my function only to summation.

However, your point on bare visitors makes sense.

latexify/constants.py
```diff
@@ -14,6 +14,15 @@
 """Latexify module-level constants"""
 
 import ast
+from typing import NamedTuple
+
+
+class Actions(NamedTuple):
```
ryosuketc, line 20: I decided to use `NamedTuple` here because of two benefits.
-   It allows dot access - e.g. `constants.actions.SET_BOUNDS` (vs. `constants.actions[key]` with a `dict` implementation)
-   Tuples are immutable.
-   `NamedTuple` can be written more simpler than a class.

I could have prepared another file `actions.py` to achieve the same interface, but `actions` are essentially constants. Thus I preferred to keep it in `constants.py`.

If you think otherwise, please let me know.

tests/regression_test.py
ryosuketc: Done.

latexify/core.py
ryosuketc: > This function can be reused with other operators (e.g., products and integrals have the same syntax). We can use some general action name rather than sum.

`set_bounds` as suggested in the new commit?

odashi (approved):
I think this change works, thanks!

PR #33, opened 2020-10-13
Adds NodeVisitorBase for dispatching visitor's behavior.

odashi:
This makes implementing multiple types of code generation easier (as #32 required).


ShigekiKarita (approved):
LGTM!

PR #36, opened 2022-09-29
Update setup.py with python_requires < 3.9 -> 3.10

voletiv:
Does it not work with Python 3.9?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ def main():
     install_requires=[
         'dill>=0.3.2',
     ],
-    python_requires='>=3.6, <3.9',
+    python_requires='>=3.6, <3.10',
   )
 
 


setup.py
```diff
@@ -45,7 +45,7 @@ def main():
     install_requires=[
         'dill>=0.3.2',
     ],
-    python_requires='>=3.6, <3.9',
+    python_requires='>=3.6, <3.10',
```
JCGoran, line 48: Seems to work fine on `3.10.6` as well, so it's probably better to just set `<3.11` (as that's the current preview version) for now.

odashi: +1. it looks good to pin 3.11 for now.

odashi:
@voletiv Thanks for proposing a PR!
I guess #39 is maybe a better fix that covers things in this PR too.

odashi:
Close this issue since the same change has been merged.

PR #39, opened 2022-10-01
Updating Required Python Version

wiljav:
Updating required Python version from 3.8 to 3.10.


wiljav:
Will not work with Python version 3.10.4

PR #40, opened 2022-10-01
Updated python version requirements

wiljav:
Updated and tested python version requirements from 3.8 to 3.10.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
 
     steps:
     - uses: actions/checkout@v2
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ def main():
     install_requires=[
         'dill>=0.3.2',
     ],
-    python_requires='>=3.6, <=3.11',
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -45,7 +48,7 @@ def main():
     install_requires=[
         'dill>=0.3.2',
     ],
-    python_requires='>=3.6, <3.9',
+    python_requires='>=3.6, <=3.11',
```
odashi, line 51: 3.11 will be released soon, but it is still in preview. I think it would be better to avoid explicit support of 3.11 for now.
```suggestion
    python_requires='>=3.6, <3.11',
```

setup.py
```diff
@@ -36,6 +36,9 @@ def main():
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.11',
```
odashi, line 41: ditto: avoid this line.
```suggestion
```

odashi (commented):
Thanks for contributing!

ShigekiKarita:
Hi, thanks for contributing.

Can you add related versions in our CI?
https://github.com/google/latexify_py/blob/30e3a7c0721466c53e00f34203940d8c7b1c4e13/.github/workflows/python-package.yml#L14

.github/workflows/python-package.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
```
odashi, line 14: Sorry for commenting multiple times.
I think we need to wrap these values by quotes. It looks the CI unexpectedly interpreted `3.10` as `3.1`.
```suggestion
        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
```

odashi (approved):

jkmackie:
Somebody apparently needs to sign the CLA.  I would like to try out latexify-py but cannot with Python 3.9.   I guess the workaround is a pip install with the --no-dependencies switch?

image

odashi:
@jkmackie It looks you tried to install the library through PyPI, but this change is not released yet. I think you need to wait for the new release, or try to directly install this branch onto your environment.

PR #44, opened 2022-10-05
Updated setuptools version from 0.0.7 to 0.0.8

wiljav:
Causing [PyPi](https://pypi.org/project/latexify-py/#history) to use the older version of this package (0.0.7) that doesn't support Python 3.10.


odashi (approved):
Thanks!

PR #49, opened 2022-10-12
Support basic math NumPy functions

Raphael-CV:
Support basic math NumPy functions


odashi (approved):
Thanks! It looks the title of this pull request seems to be "support NumPy functions" because this change does really not support the whole stuff in NumPy.

Raphael-CV:
It seems indeed that it is a better idea. I revert the last commits and will propose several pull requests.

odashi:
@Raphael-CV Thanks a lot!

PR #55, opened 2022-10-20
Fix CI

odashi:
This change fixes follows:

- Black was not working as expected since `--check` option was missing.
- Disable `E203` error in flake8 since it is not compatible with Black.
- Rename `.github/python-package.yml` to `.github/ci.yml`


ShigekiKarita (approved):
Good catch!

odashi:
Lol some of my email still don't has CLA, fixing it

PR #58, opened 2022-10-21
Refactoring of the library

odashi:
This change introduces several refactoring:

- Give more consistent argument names.
- Separate `core.py` into several modules:
  - `math_symbols.py` for math symbol manipulation.
  - `latexify_visitor.py` for the core NodeVisitor.
  - `frontend.py` for frontend interfaces.
- And many other cherrypicks.


src/latexify/math_symbols.py
```diff
@@ -0,0 +1,83 @@
...
+    "varrho",
+    "varsigma",
+    "vartheta",
+    "xi",
+    "zeta",
+    "Delta",
+    "Gamma",
+    "Lambda",
+    "Omega",
+    "Phi",
```
ShigekiKarita, line 45: nit: add \Psi?

ShigekiKarita (approved):
Nice to see well-typed and smaller modules.

src/latexify/math_symbols.py
odashi: Ah good catch up, thanks!

PR #61, opened 2022-10-22
Improve visit_Compare implementation

odashi:
This change introduces following changes:

- Add support for `is not`, `in`, `not in` operators. This includes cherrypicks of #59.
- Change the generated syntax from `a=b`, `a\op b` to `{a = b}`, `{a \op b}`.
- Support multiple comparators: `a == b == c` to `{a = b = c}`.

### DIFF
--- a/src/latexify/latexify_visitor.py
+++ b/src/latexify/latexify_visitor.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import ast
+from typing import ClassVar
 
 from latexify import constants
 from latexify import math_symbols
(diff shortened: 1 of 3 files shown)


src/latexify/latexify_visitor.py
```diff
@@ -246,27 +247,27 @@ def _wrap(child):
...
-        if isinstance(node.ops[0], ast.LtE):
-            return lstr + r"\le " + rstr
-        if isinstance(node.ops[0], ast.NotEq):
-            return lstr + r"\ne " + rstr
-        if isinstance(node.ops[0], ast.Is):
-            return lstr + r"\equiv" + rstr
-
-        return r"\mathrm{unknown\_comparator}(" + lstr + ", " + rstr + ")"
+        if len(node.ops) != 1:
+            raise SyntaxError("Multiple compares are not supported.")
```
ShigekiKarita, line 266: (optional) I think NotImplementedError is slightly better than SyntaxError if we will support the multiple compares in the future.

ShigekiKarita (approved):
LGTM

src/latexify/latexify_visitor.py
odashi: Yeah thanks. I also noticed multiple compares can be easily supported, will fix it in this PR.

PR #62, opened 2022-10-23
Improve visit_BoolOp implementation

odashi:
# Overview

Implements following changes for `BoolOp` manipulation:

- Change the generated syntax to `{x op x op ... op x}`
- Support multiple operands.

# Details

This change implements a similar strategy in `visit_Comare` in #61. Since `BoolOp` takes the same operator for every operand, `visit_BoolOp` looks relatively simpler than `visit_Compare`.

# References

- #61 

# Blocked by

- #61


ShigekiKarita (approved):
Very nice improvement

PR #64, opened 2022-10-24
Support substitution of identifiers

### LINKED ISSUE #50: Support substitutions of names
According to: https://github.com/google/latexify_py/issues/46#issuecomment-1278292646

Currently, latexify uses only the syntax information of the functions wrapped by `with_latex` and the decorator does not know any bound values outside the function. Some people want to substitute identifiers in the function into its actual value. For example:

```python
# Values bound by literals
t = 42
@with_latex
def diff(y):
    return y - t

# Values bound by given variables
def wrapper(t):
    @with_latex
    def diff(y):
        return y - t

mydiff = wrapper(42)
```

In both cases the user may want to see [ $diff(y) \triangleq y - 42$ ], not [ $diff(y) \triangleq y - t$ ].

I think implementing additional parameter to `with_latex` to specify substitutions, for example:
```python
@with_latex(substitutions={"t": t})
def diff(y):
    ...
```
is a handy solution at this moment, though users are required to write some additional stuff. To implement it, we need to define a substitution rule in the AST visitor that replaces matched identifiers to given strings (or any values that can be converted into a string).

odashi:
# Overview

Adds `identifiers` option to `get_latex` and `with_latex`.
CC: @cccntu @alecglen

It works like:
無題

At this point, this feature does not support:

- direct substitution of LaTeX because of maintaining the correct structure of AST during the whole processing.
- attributes `foo.bar` because it requires recursive analysis and another syntax of rules

# Details

`identifiers` option takes `dict[str, str]`, in which each key/value represents a substitution rule of identifier names (must follow the Python's naming convention of identifiers). The rule affects against (1) function names, (2) function arguments, and (3) any identifiers in the expression.

The process is implemented as a `NodeTransformer` and is applied among AST parsing and LaTeX codegen.

# References

- Fixes #50 

# Blocked by

- NA


src/latexify/frontend.py
```diff
@@ -11,25 +11,32 @@
...
     use_math_symbols: bool = False,
     use_raw_function_name: bool = False,
-    reduce_assignments: bool = False,
 ) -> str:
     """Obtains LaTeX description from the function's source.
 
     Args:
         fn: Reference to a function to analyze.
+        identifiers: If set, the mapping to replace identifier names in the function.
+            Keys are the original names of the identifiers, and dorresponding values are
```
ShigekiKarita, line 30: typo
```suggestion
            Keys are the original names of the identifiers, and corresponding values are
```

src/latexify/transformers/identifier_replacer_test.py
```diff
@@ -0,0 +1,99 @@
...
+    source = ast.Name(id="foo", ctx=ast.Load())
+    expected = ast.Name(id="bar", ctx=ast.Load())
+    transformed = IdentifierReplacer({"foo": "bar"}).visit(source)
+    test_utils.assert_ast_equal(transformed, expected)
+
+
+def test_name_not_replaced() -> None:
+    source = ast.Name(id="foo", ctx=ast.Load())
+    expected = ast.Name(id="foo", ctx=ast.Load())
+    transformed = IdentifierReplacer({"fooo": "bar"}).visit(source)
```
ShigekiKarita, line 28: Can you also add a substr case? e.g. `transformed = IdentifierReplacer({"fo": "bar"}).visit(source)`

ShigekiKarita (approved):

src/latexify/transformers/identifier_replacer_test.py
odashi: SGTM, will do

PR #67, opened 2022-10-29
Support substitution

### LINKED ISSUE #52: support indicies
Add suppot for indicies
```
import math
import latexify

@latexify.with_latex
def solve(a):
	return a[0]

print(solve)
```

to be something like `\mathrm{solve}(a) \triangleq a_0`

odashi:
# Overview

Add codegen rules for subscription `x[i][j][...]` -> $x_{i,j,...}$.
CC: @MAKMED1337

# Details

Codegen strategy:

1. Converts nested `ast.Subscript` to a tuple of the root value and the sequence of indices.
2. Applies the codegen rule `{value}_{", ".join(indices)}`

# References

Fixes #52

# Blocked by

NA


odashi:
Test is failing due to syntax incompatibility in some versions. Fixing it

ShigekiKarita (approved):
LGTM. Maybe latexify should provide a flag to opt-out this like `use_raw_function_name` (for example `use_raw_subscript`).

PR #68, opened 2022-10-29
Remove action enum, and add Latexify specific exceptions

odashi:
# Overview

Introduces following changes:

- Removes `Actions` enum and replace them to a raw string.
- Introduces `LatexifyError` and some subclasses.
- Introduces a behavior when `LatexifiedFunction` met some `LatexifyError`s.

# Details

`Actions` is eventually used as an identical stuff with strings. Maintaining this struct is no longer necessary as some other actions in `LatexifyVisitor` are defined as raw strings.

`LatexifyError` indicates internal errors of Latexify itself that can't be recovered by users.
`with_latexify` now captures this error to prevent destroying the runtime. Jupyter works as follows:

# References

NA

# Blocked by

NA

### DIFF
--- a/src/latexify/exceptions.py
+++ b/src/latexify/exceptions.py
@@ -1,0 +1,27 @@
+"""Exceptions used in Latexify."""
+
+
+class LatexifyError(Exception):
+    """Base class of all Latexify exceptions.
+
+    Subclasses of this exception does not mean incorrect use of the library by the user,
+    but informs users that Latexify went into something worng during compiling the given
+    functions.
+    These functions are usually captured by the frontend functions (e.g., `with_latex`)
+    to prevent destroying the entire program.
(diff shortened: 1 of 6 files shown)


src/latexify/exceptions.py
```diff
@@ -0,0 +1,36 @@
+"""Exceptions used in Latexify."""
+
+
+class LatexifyError(Exception):
+    """Base class of all Latexify exceptions.
+
+    Subclasses of this exception does not mean incorrect use of the library by the user,
+    but informs users that Latexify went into something worng during compiling the given
```
ShigekiKarita, line 8: typo
```suggestion
    but informs users that Latexify went into something wrong during compiling the given
```

ShigekiKarita (approved):
LGTM

PR #71, opened 2022-10-31
More refactoring

odashi:
# Overview

- Support all constant literals.
- Digits are wrapped by `{` and `}` so that the subscript operator works: $x_42$ --> $x_{42}$
  - This would insert unnecessarily many braces at this point, need further refactoring.
- Separate parsing strategy from `get_latex` to `parser.parse_function`.
- Test utilities:
  - Make `ast_equal` stricter.
  - Add `require_at_least/most` decorator to run tests on only specific Python verisons.
- Other trivial refactoring.

# Details

NA

# References

NA

# Blocked by

NA

### DIFF
--- a/src/latexify/latexify_visitor.py
+++ b/src/latexify/latexify_visitor.py
@@ -8,6 +8,7 @@ from typing import ClassVar
 from latexify import constants
 from latexify import math_symbols
 from latexify import node_visitor_base
+from latexify import exceptions
 
 
 class LatexifyVisitor(node_visitor_base.NodeVisitorBase):
@@ -45,8 +46,10 @@ class LatexifyVisitor(node_visitor_base.NodeVisitorBase):
 
         self.assign_var = {}
 
-    def generic_visit(self, node, action):
-        return str(node)
+    def generic_visit(self, node, action) -> str:
+        raise exceptions.LatexifyNotSupportedError(
+            f"Unsupported AST: {type(node).__name__}"
+        )
 
     def visit_Module(self, node, action):  # pylint: disable=invalid-name
         return self.visit(node.body[0], "multi_lines")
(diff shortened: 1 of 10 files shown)


src/latexify/latexify_visitor.py
```diff
@@ -171,26 +171,65 @@ def _decorated_lstr_and_arg(node, callee_str, lstr):
...
-        return str(node.n)
+        Args:
+            value: A constant value.
+
+        Returns:
+            The LaTeX representation of `value`.
+        """
+        if value is None or isinstance(value, bool):
+            return r"\mathrm{" + str(value) + "}"
+        if isinstance(value, (int, float, complex)):
```
ShigekiKarita, line 196: Maybe some scientists prefer "i" to "j" ¯\\\_(ツ)\_/¯

> In contexts in which use of the letter i is ambiguous or problematic, the letter j or the Greek [ι](https://en.wikipedia.org/wiki/Iota) is sometimes used instead.[[a]](https://en.wikipedia.org/wiki/Imaginary_unit#cite_note-3) For example, in [electrical engineering](https://en.wikipedia.org/wiki/Electrical_engineering) and [control systems engineering](https://en.wikipedia.org/wiki/Control_systems_engineering), the imaginary unit is normally denoted by j instead of i, because i is commonly used to denote [electric current](https://en.wikipedia.org/wiki/Electric_current).

https://en.wikipedia.org/wiki/Imaginary_unit

ShigekiKarita (approved):
LGTM

src/latexify/latexify_visitor.py
odashi: Haha, it would be better to provide some flag to switch the appearance of this value.

PR #72, opened 2022-10-31
More and more refactoring.

### LINKED ISSUE #69: Lack of equals sign
I dont know if it is my techniques problem. When I run the sample code on Pycharm and use the converted latex code, there is no equals sign between left hand expression and right hand expression.

11b89272a66074b52982a8a284fa87f

odashi:
# Overview

- Change the substitution sign: use $=$ rather than $\triangleq$.
- Separate `reduce_assignments` behavior to `transformers/AssignmentReducer`.
- Refactoring the behavior around "sum(... for x in range(...))".
- Remove `NodeVisitorBase` implementation.
- Move `LatexifyVisitor` to `codegen/FunctionCodegen`
- Some trivial refactoring.

# Details

## Substitution sign

$\triangleq$ is available only when the `amssymb` package is loaded, which is not compatible with the bare LaTeX. This behavior often confuses users who try to compile the obtained LaTeX source by themselves.

The sign is replaced to $=$ at this point, because I thought that this sign is the most neutral choice to every user.
We may further need to provide some flags to replace this sign by users.

## Other refactoring

This change improves the modulability of the whole library.

The `visit_*` functions in the new `FunctionCodegen` visitor returns always `str`, while ad-hoc processes are moved away to the new AST transformer `AssignmentReducer` or some private functions.

`AssignmentReducer` directly modifies the AST by replacing subtrees with substitution. This could reduce some unnecessary parenthesis pairs `\left(` and `\right)` caused by ad-hoc processing in the old Visitor.

This change also removes all "actions" from the original Visitor so that we can use the standard `ast.NodeVisitor` as the base class of the codegen.

# References

- Fixes #69

# Blocked by

NA


src/latexify/codegen/function_codegen.py
```diff
@@ -42,147 +38,114 @@ def __init__(
...
-            for prefix in constants.PREFIXES:
-                if callee_str.startswith(f"{prefix}."):
-                    callee_str = callee_str[len(prefix) + 1 :]
-                    break
-
-            lstr, rstr = constants.BUILTIN_CALLEES.get(callee_str, (None, None))
-            if lstr is None:
-                lstr = r"\mathrm{" + callee_str + r"}\left("
-                rstr = r"\right)"
+        if func_str in ("sum", "prod") and isinstance(node.args[0], ast.GeneratorExp):
```
ryosuketc, line 132: sum / prod handler here LGTM.

src/integration_tests/regression_test.py
```diff
@@ -1,154 +1,217 @@
...
+        f,
+        r"\mathrm{f}(x) = {3}(x^{{2}} + x^{{2}})",
+        reduce_assignments=True,
     )
 
-    latex_with_option = (
-        r"\mathrm{f}(x) \triangleq "
-        r"{3}\left( \left( x^{{2}} \right) + \left( x^{{2}} \right) \right)"
+
+def test_reduce_assignments_with_if() -> None:
```
ShigekiKarita, line 197: so nice

ShigekiKarita (approved):
LGTM

PR #73, opened 2022-10-31
Add latexify.function/expression frontends

### LINKED ISSUE #54: Ability to get latex without getting function name
```
import latexify
def solve(a):
	return 5 * a
print(latexify.get_latex(solve, no_function=True))
```
to be `5a`, not `\mathrm{solve}(a) \triangleq 5a`

i think it's better to create another function not parameter or set by default

odashi:
# Overview

- Add `latexify.function` and `latexify.expression` frontend functions.
- Add `use_signature` option to `latexify.get_latex`
- Deprecate `latexify.with_latex`

# Details

`latexify.function` is the same with the old `latexify.with_latex`, while `latexify.expression` generates LaTeX without the function's signature:

This behavior is controlled by the difference of the default value of the `use_signature` option in `get_latex`.

Also, `function` and `expression` would be more intuitive than the old name `with_latex` to express that the decorator generates a specific form.

# References

- Fixes #54

# Blocked by

- #72


ShigekiKarita (approved):
LGTM

PR #76, opened 2022-11-04
Support more expressions in sum and prod.

odashi:
# Overview

Supports following expressions in `sum` and `prod`:

And introduces `analyzers` and `ast_utils` submodules.

# Details

- `ast_utils` provides trivial processors on AST. Currently it contains `make_constant` and `extract_int(_or_none)` functions.
- `analyzers` provides more complicated processors on AST, typically for analyzing a "snippet" in the code. Currently it contains `analyze_range` which analyzes the invocation of the `range` function and returns its information.
- `visit_Call` in `FunctionCodegen` utilizes `analyze_range` to check if the range information is available. If not (including unsupported `range` invocation), it generates a general form of `sum` or `prod` operations.

# References

NA

# Blocked by

NA

### DIFF
--- a/src/latexify/analyzers.py
+++ b/src/latexify/analyzers.py
@@ -12,7 +12,7 @@ from latexify import ast_utils, exceptions
 class RangeInfo:
     """Information of the range function."""
 
-    # Argument subtrees. These areguments could be shallow copies of the original
+    # Argument subtrees. These arguments could be shallow copies of the original
     # subtree.
     start: ast.expr
     stop: ast.expr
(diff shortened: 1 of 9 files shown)


src/latexify/analyzers.py
```diff
@@ -0,0 +1,64 @@
...
+import dataclasses
+
+from latexify import ast_utils, exceptions
+
+
+@dataclasses.dataclass(frozen=True, eq=False)
+class RangeInfo:
+    """Information of the range function."""
+
+    # Argument subtrees. These areguments could be shallow copies of the original
```
ShigekiKarita, line 15:
```suggestion
    # Argument subtrees. These arguments could be shallow copies of the original
```

ShigekiKarita (approved):
LGTM

PR #77, opened 2022-11-04
Support multi-clause comprehension in sum and prod.

odashi:
# Overview

Supports following expression:

# Details

Simply changed the `FunctionCodegen` to analyze multiple comprehensions.

# References

NA

# Blocked by

NA

PR #79, opened 2022-11-05
Supports if clause in comprehension with sum/prod.

odashi:
# Overview

Supports following expression:

# Details

Converts the `ifs` subtrees into expression and joins all conditions with `\land` ( $\land$ ).

# References

NA

# Blocked by

- #77

### DIFF
--- a/src/integration_tests/regression_test.py
+++ b/src/integration_tests/regression_test.py
@@ -89,7 +89,7 @@ def test_sum_with_limit_1arg() -> None:
         return sum(i**2 for i in range(n))
 
     latex = (
-        r"\mathrm{sum_with_limit}(n) = \sum_{i = 0}^{{n - 1}} \left({i^{{2}}}\right)"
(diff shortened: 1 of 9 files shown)


ShigekiKarita:
The code looks good to me. But I have a few questions on this notation:

> joins all conditions with \land

Do you have any references to justify that the notation is common? I couldn't find the usage of \land in https://en.wikipedia.org/wiki/Summation 

And I think braces e.g. $(i \in n) \land (i < 3)$ would make subscripts more readable and robust against operator precedence.

ShigekiKarita (approved):

odashi:
@ShigekiKarita
This would be the corresponding explanation (in Wikipedia):



> And I think braces e.g. (...formulas...) would make subscripts more readable

Yeah I also think so, let me fix it.

odashi:
Now the output looks:

ShigekiKarita:
Fair enough. Thanks!

ShigekiKarita (approved):
LGTM again.

PR #80, opened 2022-11-08
fix: sub operate cause bracket error

chunibyo-wly:
# Overview

Because plus command and sub command has same priority, this will always cause a wrong brackest results.

# Details
image

image


odashi (approved):
Thanks! This is the reasonable fix and I will merge it after you agreed the Google's CLA.

PR #85, opened 2022-11-12
Fine-grained codegen of binary operations

odashi:
# Overview

This change refactors codegen around binary operations:

- Support all operators.
- Add appropriate parenthesis for non-associative operators.
- Change $\bmod$ to $\\%$ because the former has multiple meanings.
- Some minor refactoring.

# Details

This change introduces `BinOpRule` struct, which represents the codegen rule of each operator in a unified manner. `visit_BinOp` generates the LaTeX form according to `BinOpRule` of parent and left/right operands.

# References

NA

# Blocked by

NA


ShigekiKarita (approved):
LGTM

PR #92, opened 2022-11-12
Entire refactoring of BinOp/UnaryOp/Compare/BoolOp

odashi:
# Overview

This change contains:

- Support all operators.
- Consider operator precedences in UnaryOp/Compare/BoolOp as well as BinOp.
- Some minor refactoring.

# Details

This change introduces the operator precedences corresponding to all operator types, and removes extra parentheses appeared in several results. I believe this change introduces almost complete codegen for all operators.

The `not` operator is treated to have a higher precedence than the Python's syntax because the LaTeX counterpart ( $\lnot$ ) looks to have the same precedence with other unary operators.

# References

NA

# Blocked by

NA

### DIFF
--- a/src/latexify/codegen/function_codegen.py
+++ b/src/latexify/codegen/function_codegen.py
@@ -12,28 +12,95 @@ from latexify import math_symbols
 from latexify import exceptions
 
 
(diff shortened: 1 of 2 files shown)


ShigekiKarita (approved):
LGTM

Issue #56, opened 2022-10-20
Supporting Set operations and comprehensions


Takashiidobe:
It would be nice to support some more set operations and comprehensions, since they're pretty easy to use in python. (I assume this is going to be fairly difficult though, since it requires a fair amount of information from the python ast)

For example, this code:

```py
@latexify.with_latex
def solve():
    return {x ** 2 for x in range(1, 10)}
```

Returns this currently:

```tex
\displaystyle \mathrm{solve}() \triangleq <ast.SetComp object at 0x7f6920238160>
```

It would be nice to show something like this:

```tex
\displaystyle \mathrm{solve}() \triangleq \{x\mid x^2\in\Bbb \(1..10\)\}
```
$$ \displaystyle \mathrm{solve}() \triangleq \{x\mid x^2\in\Bbb \(1..10\)\} $$

And operations on sets:

```py
@latexify.with_latex
def solve(x: set, y: set):
    return x | y
    
```

Returns this currently:

```tex
\displaystyle \mathrm{solve}(x, y) \triangleq \mathrm{unknown\_binop}(x, y)
```

It could be something like this:

```tex
\displaystyle \mathrm{solve}(x, y) \triangleq x \cup y
```

$$ \displaystyle \mathrm{solve}(x, y) \triangleq  x \cup y $$

`in` and `not in` could also be supported, like `x in A` or `x not in A` like this:

```tex
x\in A, x\notin A
```
$$ \displaystyle \mathrm{solve}() \triangleq x\in A $$

$$ \displaystyle \mathrm{solve}() \triangleq x\notin A $$

odashi:
Tahnks! I think this is reasonable proposal.

To clarify, this includes 3 distinct features as follows:

- Support comprehension: `{f(x) for x in X}` -> $\\{f(x) | x \in X\\}$
- Support binary operators on sets: `x | y` -> $x \cup y$
- Support affiliation operators: `x in y` -> $x \in y$

We need to take care about the second case because set operations shares the same syntax with bit operations and it is not essentially distinguishable by the syntax tree. We may need something like:

- Add argument to tell which set/bit operations are preferred
- Rely on typing information (this is stretchy, not applicable for now)

odashi:
I will work on this issue this week. As for comprehension, we can support it similarly to that of `sum/prod` operations. Since binary/set operations are not trivial to distinguish, it may be good to provide a flag to switch these operators by users.

odashi:
[...]

PR #94, opened 2022-11-13
Support set operations.

odashi:
# Overview

Supports following expressions:

無題

# Details

Add another typesets of BinOp and Compare for set operations, and switch it by the flag `use_set_symbols`. This flag is disabled by default.

Since the parser can't distinguish whether the expression represents a set or not, this flag affects the entire expression in the wrapped function.

# References

- #56

# Blocked by

NA


ShigekiKarita (approved):
LGTM

PR #99, opened 2022-11-13
ci: Upgrade GitHub Actions and Python

cclauss:
# Overview

https://docs.python.org/3/whatsnew/3.11.html. Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite. See [Faster CPython](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-faster-cpython) for details.

# Details

# References

# Blocked by

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,11 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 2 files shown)


odashi:
@cclauss Thank you! I will merge this PR after one planned release (0.2) was delivered.

cclauss:
ERROR: Package 'latexify-py' requires a different Python: 3.11.0 not in '<3.11,>=3.7'

odashi (approved):
Note to @odashi : This change will also introduce Python 3.11 support.

PR #102, opened 2022-11-14
[add] configuration

chunibyo-wly:
But I don't know why we need to freeze a config object?
It's looked weired forcing use *merge* to update configuration?


chunibyo-wly:
itegration tests has failed, looks like it's about trailing space

Issue #66, opened 2022-10-25
Config class


odashi:
Follows #65

It may be good if we provide a config class integrating every setting, which are currently passed directly to `with_latex`:

```python
import latexify

config = latexify.Config.defaults()
config.use_math_symbols()
config.expand_function("expit")

@latexify.with_latex(config)
def f(x):
    return expit(x)

# Will generate: \mathrm{f}(x) \triangleq \frac{1}{1+\exp{(-x)}}
```

odashi:
possible logic:

```python
inner_config = copy(config if config is not None else Config.defaults())
for each kwarg:
    inner_config.update(kwarg)

inner_process(fn, inner_config)
```

chunibyo-wly:
working on this

odashi:
@chunibyo-wly
Thanks for working on this. Since this feature should be carefully designed, could you provide your idea about the implementation before throwing PRs?

chunibyo-wly:
hello, thanks for your reply, could you please give me some time to do some tests?
I will give you answer later

odashi:
@chunibyo-wly Sure. I targeted this feature on the release after the next (0.3) so we don't need to rush at this point.

chunibyo-wly:
1. I think we could use bitwise and dynamic method creation to do this
```python
from __future__ import annotations

from latexify.constants import CONFIGURATION
from typing import Callable

class Config:
    def __init__(self,
        identifiers: dict[str, str] | None = None,
        reduce_assignments: bool = False,
        use_math_symbols: bool = False,
        use_raw_function_name: bool = False,
        use_signature: bool = True,) -> None:

        self.flag: int = reduce_assignments * CONFIGURATION['reduce_assignments'] \
                    + use_math_symbols * CONFIGURATION["use_math_symbols"] \
                    + use_raw_function_name * CONFIGURATION["use_raw_function_name"] \
                    + use_signature * CONFIGURATION["use_signature"]
        self.identifiers: dict[str, str] = identifiers
        self.expanded_function: set[str] = set()

        for _config in CONFIGURATION:
            setattr(self, _config, self.__make_method_change_flag(_config))
            setattr(self, f"is_{_config}", self.__make_method_use_flag(_config))

    @staticmethod
    def defaults() -> Config:
        return Config()

    def expand_function(self, function: str | list[str]) -> None:
        if type(function) == str:
            self.expanded_function.add(function)
        elif type(function) == list:
            self.expand_function.update(function)

[...]

PR #103, opened 2022-11-14
[add] config

chunibyo-wly:
But I don't know why we need to freeze a config object?
It's looked weired forcing use merge to update configuration?


odashi:
Ref: #66 

@chunibyo-wly Thanks!

Just a comment about "why we need to freeze": it is basically because immutability keeps ease of maintainability and reproducibility.
Always generating a fresh struct keeps the entire library unchaged, i.e., we can guarantee `latexify.function` has no side effect and we could get exactly the same result every time we called `latexify.function` even it is invoked from multiple threads. Immutability is also applied in the entire process of this library, especially tree transformation chains in `get_latex`: all transformers always return a new AST (with references to possibly old subtrees for efficiency).

We may add some functionality to override the "global" config in the future. This kind of functionality should be explicitly mentioned that "invoking this method causes side effect and not thread safe."

src/latexify/config.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
```
odashi, line 1:
```suggestion
"""Definition of the Config class."""

from __future__ import annotations

```

src/latexify/config.py
```diff
@@ -0,0 +1,46 @@
...
+import dataclasses
+
+from typing import Any
+
+
+@dataclasses.dataclass(frozen=True)
+class Config:
+    identifiers: dict[str, str] | None
+    expanded_functions: set[str] | None
```
odashi, line 10: This is not an available config at this point.
```suggestion
```

src/latexify/config.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+import dataclasses
+
+from typing import Any
+
+
+@dataclasses.dataclass(frozen=True)
+class Config:
```
odashi, line 8:
```suggestion
class Config:
   """Configurations to control the behavior of latexify."""

```

src/latexify/config.py
```diff
@@ -0,0 +1,46 @@
...
+    identifiers: dict[str, str] | None
+    expanded_functions: set[str] | None
+    reduce_assignments: bool
+    use_math_symbols: bool
+    use_raw_function_name: bool
+    use_signature: bool
+    use_set_symbols: bool
+
+    def merge(self, *, config: Config | None = None, **kwargs) -> Config:
```
odashi, line 17: Add docstring for all public functions. It is required to follow the Google's style: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

src/latexify/config.py
```diff
@@ -0,0 +1,46 @@
...
+            return arg
+
+        return Config(**{f.name: merge_field(f.name) for f in dataclasses.fields(self)})
+
+    def expand_function(self, fun: str | list[str]) -> None:
+        if type(fun) == str:
+            self.expanded_functions.add(fun)
+        elif type(fun) == list:
+            self.expanded_functions.update(fun)
+
```
odashi, line 35: remove this method.
```suggestion
```

src/latexify/config.py
```diff
@@ -0,0 +1,46 @@
...
+    @staticmethod
+    def defaults() -> Config:
+        return Config(
+            identifiers=None,
+            expanded_functions=set(),
```
odashi, line 40:
```suggestion
```

src/latexify/frontend.py
```diff
@@ -10,18 +10,10 @@
...
-    reduce_assignments: bool = False,
-    use_math_symbols: bool = False,
-    use_raw_function_name: bool = False,
-    use_signature: bool = True,
-    use_set_symbols: bool = False,
-) -> str:
+def get_latex(fn: Callable[..., Any], *, config: Config | None = None, **kwargs) -> str:
     """Obtains LaTeX description from the function's source.
 
     Args:
```
odashi, line 19: Add appropriate "Args" section.

I also think that this section should have only descriptions about `config` and `**kwargs`. Descriptions of specific configs can be written in the `Config` class instead.

src/latexify/config.py
```diff
@@ -0,0 +1,67 @@
...
+            use_math_symbols: Whether to convert identifiers with a math symbol surface
+                (e.g., "alpha") to the LaTeX symbol (e.g., "\\alpha").
+            use_raw_function_name: Whether to keep underscores "_" in the function name,
+                or convert it to subscript.
+            use_signature: Whether to add the function signature before the expression
+                or not.
+            use_set_symbols: Whether to use set symbols or not.
+
+        Returns:
+            a new thread safe Config object
```
odashi, line 43: I think this object itself does not guarantee the thread safety.
```suggestion
            A new Config object
```

src/latexify/config.py
```diff
@@ -0,0 +1,67 @@
...
+                if config is not None:
+                    arg = getattr(config, name)
+                else:
+                    arg = getattr(self, name)
+            return arg
+
+        return Config(**{f.name: merge_field(f.name) for f in dataclasses.fields(self)})
+
+    @staticmethod
+    def defaults() -> Config:
```
odashi, line 59:
```suggestion
    def defaults() -> Config:
        """Generates a Config with default values.
        
        Returns:
            A new Config with default values
        """
```

src/latexify/frontend.py
```diff
@@ -10,58 +10,41 @@
 from latexify import exceptions
 from latexify import parser
 from latexify import transformers
+from latexify.config import Config
```
odashi, line 13: Google's style guide does not allow direct imports.

https://google.github.io/styleguide/pyguide.html#22-imports

```suggestion
from latexify import config
```

src/latexify/frontend.py
```diff
@@ -10,58 +10,41 @@
...
-        use_math_symbols: Whether to convert identifiers with a math symbol surface
-            (e.g., "alpha") to the LaTeX symbol (e.g., "\\alpha").
-        use_raw_function_name: Whether to keep underscores "_" in the function name,
-            or convert it to subscript.
-        use_signature: Whether to add the function signature before the expression or
-            not.
-        use_set_symbols: Whether to use set symbols or not.
+        config: use defined Config object, if it is None, it will be automatic assigned
+            with default value.
+        kwargs: dict of Config field values that could be defined individually by uses.
```
odashi, line 23: `kwargs` looks there is the exact parameter `kwargs`.

```suggestion
        **kwargs: dict of Config field values that could be defined individually by uses.
```

src/latexify/config.py
```diff
@@ -0,0 +1,67 @@
...
+from __future__ import annotations
+
+import dataclasses
+
+from typing import Any
+
+
+@dataclasses.dataclass(frozen=True)
+class Config:
+    """Configurations to control the behavior of latexify."""
```
odashi, line 12: Could you put descriptions of members here?

```suggestion
    """Configurations to control the behavior of latexify.
    
    Attributes:
        foo: ...
        bar: ...
    """
```

src/latexify/config.py
```diff
@@ -0,0 +1,67 @@
...
+                ^[A-Za-z_][A-Za-z0-9_]*$
+            reduce_assignments: If True, assignment statements are used to synthesize
+                the final expression.
+            use_math_symbols: Whether to convert identifiers with a math symbol surface
+                (e.g., "alpha") to the LaTeX symbol (e.g., "\\alpha").
+            use_raw_function_name: Whether to keep underscores "_" in the function name,
+                or convert it to subscript.
+            use_signature: Whether to add the function signature before the expression
+                or not.
+            use_set_symbols: Whether to use set symbols or not.
```
odashi, line 40:
```suggestion
            **kwargs: Members to modify. This value precedes both self and config.
```

odashi (commented):
Almost fine, and there are only trivial comments remaining.

odashi:
By the way, there are several pull requests that may affect config so there would be some collisions when merging. I guess it is better to merge this pull request first, and request others to merge `main` after merging this one.

odashi:
@chunibyo-wly Hi, I like this change, and would like to merge it soon! If you could resolve the comments above, I really appreciate it.

chunibyo-wly:
sorry, I forgot to check my email

src/latexify/frontend.py
chunibyo-wly: sorry, I don't know how to modify this one.
Could you help me?

ZibingZhang: you'll have to change the import, and change any uses of `Config` to `config.Config`

chunibyo-wly:
```python
def get_latex(
    fn: Callable[..., Any], *, _config: config.Config | None = None, **kwargs
) -> str:
```

Thanks for your reply. Is it looks like this one?

ZibingZhang: Ah yeah that makes it a bit awkward for the keyword argument doesn't it.

Maybe change the import to `from latexify import config as cfg`?

That way we can keep the name of the keyword argument as `..., config: cfg.Config | None = None, ...`.

chunibyo-wly: Thanks for your advice, you were right

ZibingZhang (changes requested):
Since https://github.com/google/latexify_py/pull/125 has been merged in, there's a new argument `expand_functions` in `get_latex`.

Could you update the `Config` class to include that as well?

odashi:
@ZibingZhang Yeah, but it could be processed after merging this PR.

odashi (approved):
I'd approve this. maybe we need a follow-up PR to introduce a new flag (or directly applying it here)

odashi:
I applied a monkey patch to remain `expand_functions` as is to merge this PR. After fixing the linter error I will merge it.

odashi:
@chunibyo-wly Thanks for your work! I think this is very important change for this repository.

Issue #91, opened 2022-11-12
Support IfExp


odashi:
Support the expression syntax `a if x else b`. Since we have already a codegen for if-statements, `IfExp` should generate the same LaTeX.

<img width="398" alt="無題" src="https://user-images.githubusercontent.com/1023695/201498093-813fc9a7-1d30-4796-8539-812486c0d3bc.png">

odashi:
Resolved by #111

PR #111, opened 2022-11-15
Add support for ast.IfExp

kshxtij:
# Overview

Add support for `a if x else b` (ast.IfExp) statements.

# Details

Screenshot 2022-11-15 at 13 54 16

# References

- #91 

# Blocked by

NA

### DIFF
--- a/src/latexify/codegen/function_codegen.py
+++ b/src/latexify/codegen/function_codegen.py
@@ -517,15 +517,17 @@ class FunctionCodegen(ast.NodeVisitor):
         latex += self.visit(node)
         return latex + r", & \mathrm{otherwise} \end{array} \right."
 
-    def visit_IfExp(self, node: ast.IfExp) -> Any:
+    def visit_IfExp(self, node: ast.IfExp) -> str:
         """Visit an ifexp node"""
         latex = r"\left\{ \begin{array}{ll} "
-        cond_latex = self.visit(node.test)
-        true_latex = self.visit(node.body)
-        false_latex = self.visit(node.orelse)
-        latex += true_latex + r", & \mathrm{if} \ " + cond_latex + r" \\ "
-        latex += false_latex + r", & \mathrm{otherwise} \end{array} \right."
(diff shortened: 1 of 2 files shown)


src/latexify/codegen/function_codegen.py
```diff
@@ -517,6 +517,16 @@ def visit_If(self, node: ast.If) -> str:
         latex += self.visit(node)
         return latex + r", & \mathrm{otherwise} \end{array} \right."
 
+    def visit_IfExp(self, node: ast.IfExp) -> Any:
+        """Visit an ifexp node"""
+        latex = r"\left\{ \begin{array}{ll} "
+        cond_latex = self.visit(node.test)
+        true_latex = self.visit(node.body)
+        false_latex = self.visit(node.orelse)
```
odashi, line 525: Note: I can make some follow-up pull requests for this comment.

I think this line does not process the recursion. e.g.:

```python
if a:
    return x
elif b:
    return y
else:
    return z
```

and

```python
return x if a else (y if b else z)
```

should express the same logic, but this impl may generates different result.

odashi (commented):
Hi, this is really good! Could you prepare some unit tests in `function_codegen_test.py` as well?

src/latexify/codegen/function_codegen.py
kshxtij: Screenshot 2022-11-15 at 21 24 42

I was able to process the recursion you highlighted, but wanted to know if we need to support recursion within both code blocks of the `if  then  else` statement like shown in the 2nd example.

odashi: In `visit_If` the current implementation processes only `orelse` subtree. I think we could choose the same behavior at this point.

odashi (approved):

odashi:
Great work, thanks!

odashi:
Btw, CIs (Black and flake8) are still failing, I will merge this PR after resolving these issues.

Issue #88, opened 2022-11-12
Automatically detecting imported libraries


odashi:
Ref: #86 

Sometimes users import libraries with custom names, and the current implementation of latexify does not recognize it without passing the custom config (this is not the current feature: tracked by #87)

We can anyway list up the all imported modules through `globals()`, and check if they contains some well known libraries. For example, if the user invoked `import numpy as np`, `globals()["np"]` should point to a module with `__name__ == "numpy"`.

This is clever enough I think, but applying this involves environment-dependent behavior. I think this can be implemented as an additional NodeTransformer with a boolean switch which turns off by default.

ZibingZhang:
Once this happens, I think we should get rid of `_COMMON_PREFIXES`. The trimming of prefixes should be by default turned off for all imo.

odashi:
> Once this happens, I think we should get rid of _COMMON_PREFIXES

Though prefixes trimming is a weird behavior, it couldn't be completely replaced by this feature. We couldn't expect that appropriate libraries are imported when `@latexify.function` is applied, because function can be defined with undefined symbols.

Btw, I guessed we already don't need prefix trimming of especially `math`, `numpy` and `np` because we have attribute-agnostic way to extract function names.

https://github.com/google/latexify_py/blob/ddcbed9cb2105c67b5820d079038dd220559f1b9/src/latexify/ast_utils.py#L146

odashi:
Tried to remove common prefixes, observed that only this test failed. It is fine to remove this behavior I guess (I will work)

<img width="876" alt="無題" src="https://user-images.githubusercontent.com/1023695/206904381-39e2f74b-0f96-4aba-9a5d-8fbbd1e518e7.png">

odashi:
Maybe this is a weird behavior though:

<img width="337" alt="無題" src="https://user-images.githubusercontent.com/1023695/206904578-2cb870aa-fb10-4831-9514-9a47c8fdd946.png">

ZibingZhang:
> Though prefixes trimming is a weird behavior, it couldn't be completely replaced by this feature. We couldn't expect that appropriate libraries are imported when `@latexify.function` is applied, because function can be defined with undefined symbols.

Yeah we should definitely keep the prefix trimming feature, just make the user specify which prefixes instead of defaulting to a hardcoded list.

> Maybe this is a weird behavior though

That's definitely odd haha. Should maybe only convert in `visit_Name` and not `visit_Attribute`?

odashi:
[...]

Issue #87, opened 2022-11-12
Custom list of prefixes that should be trimmed


odashi:
Ref: #86 

Currently we hard-coded function prefixes to be trimmed as follows:

https://github.com/google/latexify_py/blob/86e612963ee4c8d92c6812f99dd75e2309bc9735/src/latexify/constants.py#L5

It would be good if the users can set it manually.

odashi:
This should be an additional features of #81.

Issue #81, opened 2022-11-09
Package name trimming as a NodeTransformer


odashi:
Prefixes of well-known packages are trimmed by the following code block in `function_codegen.py`

https://github.com/google/latexify_py/blob/071a12d06ae67c578f92f973945c0067b3f4976d/src/latexify/codegen/function_codegen.py#L127-L132

As I noted in the TODO, this can be implemented as a NodeTransformer preprocessor. This change would make the behavior more flexible.

For example, the AST of `math.sqrt(x)` is:

```
Call(
    func=Attribute(
        value=Name(id='math', ctx=Load()),
        attr='sqrt',
        ctx=Load()),
    args=[
        Name(id='x', ctx=Load())],
    keywords=[])
```

and we can implement NodeTransformer to modify the above to:

```
Call(
    func=Name(id='sqrt', ctx=Load()),
    args=[
        Name(id='x', ctx=Load())],
    keywords=[])
```

The modification may be applied on only the `func` subtree.

Eric-Mendes:
Just found out about this lib and was playing around with it. The trimming on functions amazed me, but I noticed that package constants don't have the prefixes trimmed. Is this on purpose? Should I open an issue for it?

This is not a problem to my current work, but I thought it was a nice observation to make here.

Example:
![image](https://user-images.githubusercontent.com/42689328/202034573-881901ed-be85-411a-8028-3550608efd58.png)

odashi:
@Eric-Mendes Thanks!

Ty first thought about the strategy is as follows:

1. If the parser found that an identifier involves some prefix (in your case, `np`), the parser investigates whether the prefix is pointing to a module or not.
2. If it is a module, try to obtain the real name of the module (`np` -> `numpy`)
3. Checks if the real module name is listed in the pre-defined list (e.g., `["math", "numpy", "tensorflow", "torch"]`), the parser replaces the original identifier by removing the prefix (`np.inf` -> `inf`).

I think this should also work for the case you mentioned.

Eric-Mendes:
It's been a while since my last compilers class in college, but can I try to work on this, @odashi?

Eric-Mendes:
Just tried it, @odashi 😄
I hope I'm on the right track.

PR #113, opened 2022-11-16
refactor: prefix trimming as a node transformer

Eric-Mendes:
# Overview
Implemented the change proposed in issue #81.

# Details
Couldn't figure out a nice way of implementing automatic detection of used libraries and its aliases as discussed in #88, so in this implementation the user needs to pass them in a `set[str]` called `prefixes`. 

# References
#81 #87 

# Blocked by
NA

### DIFF
--- a/src/latexify/frontend.py
+++ b/src/latexify/frontend.py
@@ -22,7 +22,7 @@ def get_latex(
     use_raw_function_name: bool = False,
     use_signature: bool = True,
     use_set_symbols: bool = False,
-    prefixes: set[str] = set(),
+    prefixes: set[str] | None = None,
 ) -> str:
     """Obtains LaTeX description from the function's source.
 
@@ -59,9 +59,10 @@ def get_latex(
         tree = transformers.IdentifierReplacer(identifiers).visit(tree)
     if reduce_assignments:
         tree = transformers.AssignmentReducer().visit(tree)
-    # assume that both prefixes and PREFIXES are sets.
     # merge and de-duplicate the prefix list.
-    merged_prefixes = set(constants.PREFIXES) | prefixes
+    if prefixes is None:
+        prefixes = set()
+    merged_prefixes = constants.PREFIXES | prefixes
     tree = transformers.PrefixTrimmer(merged_prefixes).visit(tree)
 
     # Generates LaTeX.
--- a/src/latexify/frontend.py
+++ b/src/latexify/frontend.py
@@ -21,6 +21,7 @@ def get_latex(
     use_raw_function_name: bool = False,
     use_signature: bool = True,
     use_set_symbols: bool = False,
+    prefixes: list[str] | None = None,
 ) -> str:
     """Obtains LaTeX description from the function's source.
 
@@ -40,9 +41,11 @@ def get_latex(
         use_signature: Whether to add the function signature before the expression or
             not.
         use_set_symbols: Whether to use set symbols or not.
-
+        prefixes: Package prefixes to trim. If an alias is used, it should be in the list,
+            instead of the package itself. Example: for an `import numpy as np`, you pass
+            `prefixes=["np"]`
     Returns:
-        Generatee LaTeX description.
+        Generated LaTeX description.
 
     Raises:
         latexify.exceptions.LatexifyError: Something went wrong during conversion.
@@ -55,6 +58,8 @@ def get_latex(
         tree = transformers.IdentifierReplacer(identifiers).visit(tree)
     if reduce_assignments:
         tree = transformers.AssignmentReducer().visit(tree)
+    if prefixes is not None:
+        tree = transformers.PrefixTrimmer(prefixes).visit(tree)
 
     # Generates LaTeX.
     return codegen.FunctionCodegen(
--- a/src/latexify/frontend.py
+++ b/src/latexify/frontend.py
@@ -10,6 +10,7 @@ from latexify import codegen
 from latexify import exceptions
 from latexify import parser
 from latexify import transformers
+from latexify import constants
 
 
 def get_latex(
@@ -21,7 +22,7 @@ def get_latex(
     use_raw_function_name: bool = False,
     use_signature: bool = True,
     use_set_symbols: bool = False,
-    prefixes: list[str] | None = None,
+    prefixes: set[str] = set(),
 ) -> str:
     """Obtains LaTeX description from the function's source.
 
@@ -41,9 +42,9 @@ def get_latex(
         use_signature: Whether to add the function signature before the expression or
             not.
         use_set_symbols: Whether to use set symbols or not.
-        prefixes: Package prefixes to trim. If an alias is used, it should be in the list,
+        prefixes: Package prefixes to trim. If an alias is used, it should be in the set,
             instead of the package itself. Example: for an `import numpy as np`, you pass
-            `prefixes=["np"]`
+            `prefixes={"np"}`
     Returns:
         Generated LaTeX description.
 
@@ -60,7 +61,7 @@ def get_latex(
         tree = transformers.AssignmentReducer().visit(tree)
     # assume that both prefixes and PREFIXES are sets.
     # merge and de-duplicate the prefix list.
-    merged_prefixes = constants.PREFIXES | prefixes
+    merged_prefixes = set(constants.PREFIXES) | prefixes
     tree = transformers.PrefixTrimmer(merged_prefixes).visit(tree)
 
     # Generates LaTeX.
--- a/src/latexify/frontend.py
+++ b/src/latexify/frontend.py
@@ -58,8 +58,10 @@ def get_latex(
         tree = transformers.IdentifierReplacer(identifiers).visit(tree)
     if reduce_assignments:
         tree = transformers.AssignmentReducer().visit(tree)
-    if prefixes is not None:
-        tree = transformers.PrefixTrimmer(prefixes).visit(tree)
+    # assume that both prefixes and PREFIXES are sets.
+    # merge and de-duplicate the prefix list.
+    merged_prefixes = constants.PREFIXES | prefixes
+    tree = transformers.PrefixTrimmer(merged_prefixes).visit(tree)
 
     # Generates LaTeX.
     return codegen.FunctionCodegen(
--- a/src/latexify/transformers/prefix_trimmer.py
+++ b/src/latexify/transformers/prefix_trimmer.py
@@ -9,8 +9,6 @@ class PrefixTrimmer(ast.NodeTransformer):
     """NodeTransformer to trim prefixes of identifiers.
 
     Example:
-        import math
-
         def foo(n):
             return math.sqrt(n)
 
--- a/src/latexify/transformers/prefix_trimmer.py
+++ b/src/latexify/transformers/prefix_trimmer.py
@@ -27,7 +27,7 @@ class PrefixTrimmer(ast.NodeTransformer):
         """Initializer.
 
         Args:
-            prefixes: User defined prefixes.
+            prefixes: Set of prefixes to be trimmed.
         """
         self._prefixes = prefixes
 
--- a/src/latexify/transformers/prefix_trimmer.py
+++ b/src/latexify/transformers/prefix_trimmer.py
@@ -12,14 +12,14 @@ class PrefixTrimmer(ast.NodeTransformer):
         def foo(n):
             return math.sqrt(n)
 
-        PrefixTrimmer(["math"]) will modify the AST of
+        PrefixTrimmer({"math"}) will modify the AST of
         the function above to below:
 
         def foo(n):
             return sqrt(n)
     """
 
-    def __init__(self, prefixes: list[str]):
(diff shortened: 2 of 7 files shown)


src/latexify/frontend.py
```diff
@@ -55,6 +58,8 @@ def get_latex(
         tree = transformers.IdentifierReplacer(identifiers).visit(tree)
     if reduce_assignments:
         tree = transformers.AssignmentReducer().visit(tree)
+    if prefixes is not None:
+        tree = transformers.PrefixTrimmer(prefixes).visit(tree)
```
odashi, line 62: We have a pre-defined list of prefixes: 

https://github.com/google/latexify_py/blob/3b1b05bc5d6d9811799327dc268001e94bf68a28/src/latexify/constants.py#L5

This list is used in `FunctionCodegen` to trim the prefix, but we can employ this list for the input of `PrefixTrimmer` as well, and remove the ad-hoc implementation in `FunctionCodegen`.

```suggestion
    # assume that both prefixes and PREFIXES are sets.
    # merge and de-duplicate the prefix list.
    merged_prefixes = constants.PREFIXES | prefixes
    tree = transformers.PrefixTrimmer(merged_prefixes).visit(tree)
```

src/latexify/transformers/prefix_trimmer.py
```diff
@@ -0,0 +1,43 @@
+"""Transformer to trim package prefixes."""
+
+from __future__ import annotations
+
+import ast
+
+
+class PrefixTrimmer(ast.NodeTransformer):
+    """NodeTransformer to trim package prefixes.
```
odashi, line 9: This transformer trims not only packages, but also any identifiers with the attribute syntax.

```suggestion
    """NodeTransformer to trim prefixes of identifiers.
```

src/latexify/transformers/prefix_trimmer.py
```diff
@@ -0,0 +1,43 @@
...
+        import math
+
+        def foo(n):
+            return math.sqrt(n)
+
+        PrefixTrimmer(["math"]) will modify the AST of
+        the function above to below:
+
+        import math
+
```
odashi, line 21: ditto
```suggestion
```

src/latexify/transformers/prefix_trimmer.py
```diff
@@ -0,0 +1,43 @@
...
+import ast
+
+
+class PrefixTrimmer(ast.NodeTransformer):
+    """NodeTransformer to trim package prefixes.
+
+    Example:
+        import math
+
```
odashi, line 13: Import statement is not necessary to parse the function.
```suggestion
```

src/latexify/transformers/prefix_trimmer.py
```diff
@@ -0,0 +1,43 @@
...
+        def foo(n):
+            return sqrt(n)
+    """
+
+    def __init__(self, prefixes: list[str]):
+        """Initializer.
+
+        Args:
+            prefixes: User defined prefixes.
```
odashi, line 30:
```suggestion
            prefixes: Set of prefixes to be trimmed.
```

src/latexify/transformers/prefix_trimmer.py
```diff
@@ -0,0 +1,43 @@
...
+    def visit_Call(self, node: ast.Call) -> ast.Name:
+        """Visitor of Call."""
+        if (
+            isinstance(node.func, ast.Attribute)
+            and node.func.value.id in self._prefixes
+        ):
+            new_node_func = ast.Name(id=node.func.attr, ctx=node.func.ctx)
+            node.func = new_node_func
+
+        return node
```
odashi, line 43: Some comments:

1. This process does not handle nested attributes correctly, e.g., analyzing `foo.bar.baz` will produce a wrong `func`. I think we have to take either:
   - Process only the `foo.bar` syntax: check the `node.func.attr` is the exactly the `Name` node, and if not, skip processing it.
   - Support any number of attribute chains. (I think this is much desirable)
2. Avoid updating the original node, create a new `Call` node with replaced/original attributes instead. Since the original node could be a part of other ASTs, modifying it may cause unexpected side effects in other routines.
3. Every subtree (= every attribute of `Call` that inherits `ast.AST`) must be processed recursively.
4. `ctx` should be set as that of the original `Name` node, not the `Attribute`.

src/latexify/transformers/prefix_trimmer_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+from __future__ import annotations
+
+import ast
+
+from latexify import test_utils
+from latexify.transformers.prefix_trimmer import PrefixTrimmer
+
+
+def test_correct_call_trim() -> None:
```
odashi, line 11: Add test cases with:

1. Attributes that `PrefixTrimmer` doesn't trim.
2. Nested attributes: `foo.bar.baz`.

src/latexify/transformers/prefix_trimmer.py
odashi: I think we don't need to synthesize the new `Name`; it is enough to use the underlying `Name` in the original tree.

src/latexify/frontend.py
```diff
@@ -21,6 +21,7 @@ def get_latex(
     use_raw_function_name: bool = False,
     use_signature: bool = True,
     use_set_symbols: bool = False,
+    prefixes: list[str] | None = None,
```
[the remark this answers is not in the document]
Eric-Mendes: Agreed. Will fix that on further commits 👍🏻

src/latexify/transformers/prefix_trimmer.py
Eric-Mendes: > I think we don't need to synthesize the new `Name`; it is enough to use the underlying `Name` in the original tree.

Don't think I quite understood the 4th comment, neither the one I mentioned above. Could you elaborate further?

Sorry for the mess, it's my first open source contribution 😄

odashi: Great, I think you implemented the code very quickly, and it is much clean! Sometimes I needed to point out style issues during reviewing, but in this request I could focus on the algorithm itself, which is really nice.

As for the last suggestion, it is followed by the next comment: we can re-use the original `Name`  at the end of the attribute chain in the original tree.
For example, the AST of `foo.bar` looks like:
```python
func = Attribute(
  value='foo',
  attr=Name(id='bar', ctx=something)
)
```
and the node `func.attr` holds the exact information about the new `func`. So we can simply replace the whole `func` with `func.attr`.

Eric-Mendes: In fact the ast of `foo.bar` looks like this, @odashi:
```python
>> import ast
>> print(ast.dump(ast.parse("foo.bar", mode='eval').body))

Attribute(value=Name(id='foo', ctx=Load()), attr='bar', ctx=Load())
```
And actually, from my tests with the code above, `Attribute.attr` will always be a `str`, that's why I was leaning towards instancing a new `Name` object. 

Am I missing something here?

Eric-Mendes: Just commited new code fixing some stuff we talked about on these conversations. I believe it's a bit better, but tests are still failing and I have no idea why. See if you can spot anything.

odashi:
@Eric-Mendes If you finished refactoring, please push the re-request button (the arrow circle) on the right pane.

src/latexify/frontend.py
```diff
@@ -21,6 +22,7 @@ def get_latex(
     use_raw_function_name: bool = False,
     use_signature: bool = True,
     use_set_symbols: bool = False,
+    prefixes: set[str] = set(),
```
odashi, line 25: Mutable objects can't be used as the default values ([pylint-W0102](https://vald-phoenix.github.io/pylint-errors/plerr/errors/basic/W0102.html)). Use None instead and manually check if the argument is a set or None.

The following example shows why it is dangerous:
```
>>> def f(a=set()):
...     a.add(len(a))
...     return a
... 
>>> f()
{0}
>>> f()
{0, 1}
>>> f()
{0, 1, 2}
```

```suggestion
    prefixes: set[str] | None = None,
```

src/latexify/frontend.py
```diff
@@ -55,6 +59,10 @@ def get_latex(
         tree = transformers.IdentifierReplacer(identifiers).visit(tree)
     if reduce_assignments:
         tree = transformers.AssignmentReducer().visit(tree)
+    # assume that both prefixes and PREFIXES are sets.
+    # merge and de-duplicate the prefix list.
+    merged_prefixes = set(constants.PREFIXES) | prefixes
```
odashi, line 64: Since PREFIXES is no longer referenced from somewhere else, the definition can be converted to a set.

src/latexify/frontend.py
```diff
@@ -55,6 +59,10 @@ def get_latex(
         tree = transformers.IdentifierReplacer(identifiers).visit(tree)
     if reduce_assignments:
         tree = transformers.AssignmentReducer().visit(tree)
+    # assume that both prefixes and PREFIXES are sets.
```
odashi, line 62: This is guaranteed by type hints, no longer needed
```suggestion
```

src/latexify/transformers/prefix_trimmer.py
odashi: Ah sorry I'd suggested a wrong comment. Yes you are right. The Attribute node is defined as follows:

無題

src/latexify/transformers/prefix_trimmer.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self, prefixes: set[str]):
+        """Initializer.
+
+        Args:
+            prefixes: Set of prefixes to be trimmed.
+        """
+        self._prefixes = prefixes
+
+    def visit_Call(self, node: ast.Call) -> ast.Call:
```
odashi, line 30: I think `visit_Call` is no longer necessary, as you implemented `visit_Attribute` instead, which can process every `Attribute` node in the tree.

src/latexify/transformers/prefix_trimmer.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return node
+
+    def visit_Attribute(self, node: ast.Attribute) -> ast.Name | ast.Attribute:
+        """Visitor of Attribute."""
+        node_value = node.value
+        while isinstance(node_value, ast.Attribute):
+            node_value = node_value.value
+
+        # if outermost prefix matches any given prefix, loose
```
odashi, line 53: I think this policy is not safe as the dropped subtree may involve the stuff that the users didn't expect. It would be better to either drop only the leftmost prefix, and provide a way to specify longer chains: "foo.bar" (but this may be somewhat beyond the scope of this pull request).

src/latexify/transformers/prefix_trimmer_test.py
```diff
@@ -0,0 +1,86 @@
...
+        keywords=[],
+    )
+    expected = ast.Call(
+        func=ast.Name(id="sqrt", ctx=ast.Load()),
+        args=[ast.Name(id="x", ctx=ast.Load())],
+        keywords=[],
+    )
+    transformed = PrefixTrimmer({""}).visit(source)
+    with pytest.raises(AssertionError):
+        test_utils.assert_ast_equal(transformed, expected)
```
odashi, line 64: Specify `expected` as the exact result of the `PrefixTrimmer`. The test in L63-L64 actually does not check anything (imagine that we changed `expected` to any random ASTs, then the test always passes)

src/latexify/transformers/prefix_trimmer_test.py
```diff
@@ -0,0 +1,86 @@
...
+        keywords=[],
+    )
+    expected = ast.Call(
+        func=ast.Name(id="lat", ctx=ast.Load()),
+        args=[ast.Name(id="x", ctx=ast.Load())],
+        keywords=[],
+    )
+    transformed = PrefixTrimmer({"lat"}).visit(source)
+    with pytest.raises(AssertionError):
+        test_utils.assert_ast_equal(transformed, expected)
```
odashi, line 86: ditto

odashi:
@Eric-Mendes Hi, All failed tests belong to `function_codegen_test.py`. These tests only invoke `FunctionCodegen` without any preprocessing.

Since this pull request removed the trimming behavior from `FunctionCodegen`, the codegen no longer recognize any prefixes, resulting in some test failures. you need to fix the input text of test cases to have trimmed identifiers, e.g., `math.prod` -> `prod`

src/latexify/transformers/prefix_trimmer.py
Eric-Mendes: Without it tests fail. Not sure why

odashi: Could you provide the example outputs you observed?

odashi:
@Eric-Mendes Hi, it looks the test is still failing, and there's some conflicts against main. I'd appreciate it if you could resolve these issues.

odashi:
Closes this pull request due to more than 1 week inactivity.

Issue #74, opened 2022-11-03
n+1-1 in sum/prod


odashi:
Currently the analyzer of `sum` and `prod` does not recognize the reducable addition and subtraction in the `stop` parameter of `range`:

<img width="327" src="https://user-images.githubusercontent.com/1023695/199804885-dd54386f-f5e8-477f-a04e-89c45e898796.png">

It would be good if we could get $n$ rather than $n+1-1$ in most cases. This reduction would be safe because [the operands of `range` must be integers](https://docs.python.org/3/library/stdtypes.html#range), whose objects must be always associative and commutative in terms of operations $+$ and $-$.

odashi:
Note that in general arithmetic is not associative even if all operands are numbers. For example:

<img width="442" src="https://user-images.githubusercontent.com/1023695/199812204-e9b7b4e4-5b7d-4929-8a5c-a1cc50e48ae7.png">

So this treatment is available only when the function always takes integers.

Casper-Guo:
Implemented here:

https://github.com/google/latexify_py/blob/c6c291d3bcce28a269d175be384ab955ac2100c8/src/latexify/codegen/function_codegen.py#L335-L354

Current behavior is caused by not being able to convert `n+1` to a numeric type. 

The `n+1` case can perhaps be treated as a special case since it is the most common but a more general solution might be preferred. Since reduction is only needed for `n+x` or `n-x` maybe we can explicitly check for these two cases.

odashi:
@Casper-Guo I think the code above is an obsolete permalink. In the current implementation, FunctionCodegen introduced an additional analyzer that obtains subtree of range:

https://github.com/google/latexify_py/blob/071a12d06ae67c578f92f973945c0067b3f4976d/src/latexify/codegen/function_codegen.py#L334

We can implement the specific process if we detected `range_info.start` is BinOp and the operator is Add or Sub.

Casper-Guo:
We would like to work on this. Can you please assign me and @lakeblair.

odashi:
@Casper-Guo Thanks for taking your effort! Assigned to you anyway since you are the only person who attended this discussion.

odashi:
Resolved by #115

PR #115, opened 2022-11-16
Addition / Subtraction Simplification for the stop parameter in sum / prod

Casper-Guo:
<Reducing addition and subtraction in sum and prod stop parameter>

# Overview

This PR addresses issue #74. 

# Details

## Implementation 
First check the `stop` parameter of `range_info` is `ast.BinOp` and represents an addition or subtraction. If so, the minus one is applied to `BinOp.right` and a new `ast.BinOp` node is created. Then the existing visit function is leveraged to create the needed Latex.

Note the special cases `+ 1 - 1` and `- (-1) - 1` for which only `BinOp.left` is kept.

Added test cases to `regression_test.py`. Please let me know if these are more appropriate elsewhere (e.g. `function_codegen_test.py`)

## Caveats 
- `n+1-1` can be reduced to `n` but `1+n-1` cannot. This is because `range(1+n)` is unlikely and unusual. Support for this can be easily added if desired
- chained additions and subtractions cannot be fully reduced. `n + 3+4-1` will become `n+3+3` not `n+6`. This is in line with how other chained BinOps are currently handled and may be addressed in a future issue

## Questions
The implementation is slightly verbose and perhaps should be written as a separate function. Might other future PRs need this similar functionality?

# References
Issue #74

### DIFF
--- a/src/latexify/codegen/function_codegen.py
+++ b/src/latexify/codegen/function_codegen.py
@@ -588,14 +588,18 @@ class FunctionCodegen(ast.NodeVisitor):
                         upper = "{" + self.visit(range_info.stop.left) + "}"
                     else:
                         reduced_constant = ast.Constant(range_info.stop.right.value - 1)
-                        new_node = ast.BinOp(range_info.stop.left, range_info.stop.op, reduced_constant)
-                        upper = "{" + self.visit(new_node) + "}" 
+                        new_node = ast.BinOp(
+                            range_info.stop.left, range_info.stop.op, reduced_constant
+                        )
+                        upper = "{" + self.visit(new_node) + "}"
                 else:
                     if range_info.stop.right.value == -1:
                         upper = "{" + self.visit(range_info.stop.left) + "}"
                     else:
                         reduced_constant = ast.Constant(range_info.stop.right.value + 1)
-                        new_node = ast.BinOp(range_info.stop.left, range_info.stop.op, reduced_constant)
+                        new_node = ast.BinOp(
+                            range_info.stop.left, range_info.stop.op, reduced_constant
+                        )
                         upper = "{" + self.visit(new_node) + "}"
             else:
                 upper = "{" + self.visit(range_info.stop) + " - 1}"
--- a/src/latexify/codegen/function_codegen.py
+++ b/src/latexify/codegen/function_codegen.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import ast
+import sys
 import dataclasses
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


src/latexify/codegen/function_codegen.py
```diff
@@ -558,7 +558,28 @@ def _get_sum_prod_range(self, node: ast.comprehension) -> tuple[str, str] | None
             lower_rhs = f"{{{range_info.start_int}}}"
 
         if range_info.stop_int is None:
-            upper = "{" + self.visit(range_info.stop) + " - 1}"
+            # special processing needed if range_info.stop involves addition or subtraction
+            if (
+                isinstance(range_info.stop, ast.BinOp)
+                and isinstance(range_info.stop.op, (ast.Add, ast.Sub))
```
odashi, line 564: This condition is not necessary.
```suggestion
```

src/latexify/codegen/function_codegen.py
```diff
@@ -558,7 +558,28 @@ def _get_sum_prod_range(self, node: ast.comprehension) -> tuple[str, str] | None
...
+                and isinstance(range_info.stop.right, ast.Constant)
+            ):
+                if isinstance(range_info.stop.op, ast.Add):
+                    if range_info.stop.right.value == 1:
+                        upper = "{" + self.visit(range_info.stop.left) + "}"
+                    else:
+                        reduced_constant = ast.Constant(range_info.stop.right.value - 1)
+                        new_node = ast.BinOp(range_info.stop.left, range_info.stop.op, reduced_constant)
+                        upper = "{" + self.visit(new_node) + "}" 
+                else:
```
odashi, line 574: Instead, change this `else` clause to `elif`:
```suggestion
                elif isinstance(range_info.stop.op, ast.Sub):
```

odashi (commented):
The logic looks good, thanks!
Please add appropriate unit tests into `function_codegen_test.py` as well.

src/latexify/codegen/function_codegen.py
```diff
@@ -558,7 +558,28 @@ def _get_sum_prod_range(self, node: ast.comprehension) -> tuple[str, str] | None
             lower_rhs = f"{{{range_info.start_int}}}"
 
         if range_info.stop_int is None:
-            upper = "{" + self.visit(range_info.stop) + " - 1}"
+            # special processing needed if range_info.stop involves addition or subtraction
+            if (
+                isinstance(range_info.stop, ast.BinOp)
+                and isinstance(range_info.stop.op, (ast.Add, ast.Sub))
+                and isinstance(range_info.stop.right, ast.Constant)
```
odashi, line 565: `ast.Constant` is only available from Python 3.8. in 3.7 this logic should use `ast.Num`

odashi: That means we need to implement 2 variants of this logic, which are switched by `if sys.version_info.minor < 8`

Casper-Guo:
What is the configuration for the black style check and how can I address it locally?

odashi:
@Casper-Guo Could you resolve the conflicts introduced by recent changes on `main`?

Casper-Guo:
The commit history is a bit messed up after merging, please review my commits only.

odashi:
@Casper-Guo It is not easy to select only your stuff. Try `git merge main` to remove duplicated changes.

Casper-Guo:
> @Casper-Guo It is not easy to select only your stuff. Try `git merge main` to remove duplicated changes.

I rebased on main before pushing the new changes which should conceal all the new commits... Not sure why it didn't turn out that way. I will try to figure out a way to tidy it up.

odashi (approved):

odashi:
Looks great, thanks!

Issue #78, opened 2022-11-04
NDArray support


odashi:
Sometimes users want to support NDArrays in this library, e.g., `np.array([[a, b], [c, d]])` to 

$$\left( \begin{array}{cc}a&b \\\\ c&d\end{array} \right)$$

This is definitely useful, but often it is not possible to compile:

- Large arrays.
- Arrays with more than 2 dimensions.

I think we could start implementing it with reasonable restrictions (e.g., only NDArrays with <= 2 dimensions with small lengths), but it would be also good to keep discussion.

Refs:
- #38
- #53
- #75

Casper-Guo:
For 2D arrays, limiting the maximum size to `10 x 10` is perhaps a reasonable starting point. Larger arrays, especially if they are dense, are not great as visual aids even if they are rendered into LaTex.

moringspeaker:
> For 2D arrays, limiting the maximum size to `10 x 10` is perhaps a reasonable starting point. Larger arrays, especially if they are dense, are not great as visual aids even if they are rendered into LaTex.

I think it's a great idea to constraint the maximum size of matrices. I won't use matrix larger than 10 * 10 in Latex.

odashi:
@Casper-Guo Yes I think it is reasonable. maybe we restrict to show only $(R, 1)$, $(1, C)$ and $(R, C)$ arrays with size $R \le R_{max}$ and $C \le C{max}$, where $R_{max}$ and $C_{max}$ is a config defaulted by some number (10 for example)

Casper-Guo:
What should the configuration interface look like? I feel this should be file level rather than function level. 

@lakeblair

odashi:
@Casper-Guo We can start with hard-coding the constraint into the code until we decided the structure of config files.

kshxtij:
Is there a specific way, you'd suggest I implement this?

Casper-Guo:
`np.eye`, `np.ones` and other matrix functions will probably be dealt with separately. I think the biggest issue with your current implementation is that it doesn’t generalize to 1D arrays / vectors

PR #118, opened 2022-11-16
Add support for numpy matrices

kshxtij:
# Overview

Adds support for np.ndarray and np.array

# Details

Screenshot 2022-11-16 at 18 34 58

# References

- #78

# Blocked by

NA

### DIFF
--- a/src/latexify/codegen/function_codegen.py
+++ b/src/latexify/codegen/function_codegen.py
@@ -355,6 +355,9 @@ class FunctionCodegen(ast.NodeVisitor):
             (r"\mathrm{" + func_str + r"}\left(", r"\right)"),
         )
 
+        if func_str == "ndarray" or func_str == "array":
+            return self._generate_matrix(node)
+
         if func_str in ("sum", "prod") and isinstance(node.args[0], ast.GeneratorExp):
             elt, scripts = self._get_sum_prod_info(node.args[0])
             scripts_str = [rf"\{func_str}_{{{lo}}}^{{{up}}}" for lo, up in scripts]
@@ -362,6 +365,35 @@ class FunctionCodegen(ast.NodeVisitor):
 
         arg_strs = [self.visit(arg) for arg in node.args]
         return lstr + ", ".join(arg_strs) + rstr
+
+    def _generate_matrix(self, node: ast.Call) -> str:
+        """Generates a matrix from numpy.matrix."""
+        if len(node.args) != 1:
+            # raise with function name
+            raise ValueError(
+                f"Invalid number of arguments {len(node.args)}"
+            )
+        
+        arg = node.args[0]
+        # TODO: Support string being passed to numpy.matrix.
+        if not isinstance(arg, ast.List):
+            raise exceptions.LatexifyNotSupportedError(
+                "Only matrices represented by lists are supported."
+            )
+
+        # TODO: Support other types of matrix environments like (pmatrix, matrix, Bmatrix, etc.)
+        latex = r"\begin{bmatrix}"
+        if type(arg) == ast.List:
+            for elt in arg.elts:
+                if type(elt) == ast.List:
+                    for sub_elt in elt.elts:
+                        latex += self.visit(sub_elt) + r" & "
+                    latex = latex[:-2] + r" \\ "
+                else:
+                    latex += self.visit(elt) + r" & "
+            latex = latex[:-3] + r"\end{bmatrix}"
+
+        return latex
 
     def visit_Attribute(self, node: ast.Attribute) -> str:
         vstr = self.visit(node.value)
--- a/src/latexify/codegen/function_codegen.py
+++ b/src/latexify/codegen/function_codegen.py
@@ -355,8 +355,27 @@ class FunctionCodegen(ast.NodeVisitor):
             (r"\mathrm{" + func_str + r"}\left(", r"\right)"),
         )
 
-        if func_str == "ndarray" or func_str == "array":
-            return self._generate_matrix(node)
+        if func_str in ("ndarray", "array"):
+            arg = node.args[0]
+            if not isinstance(arg, ast.List) or not arg.elts:
+                return None
+
+            row0 = node.elts[0]
+
+            if not isinstance(row0, ast.List):
+                return self._generate_ndarray([self.visit(x) for x in arg.elts])
+
+            if not row0.elts:
+                return None
+
+            nCols = len(row0.elts)
+
+            if not all(isinstance(row, ast.List) and len(row.elts) == nCols for row in arg.elts):
+                return None
+
+            return self._generate_ndarray(
+                [[self.visit(x) for x in row.elts] for row in arg.elts]
+            )
 
         if func_str in ("sum", "prod") and isinstance(node.args[0], ast.GeneratorExp):
             elt, scripts = self._get_sum_prod_info(node.args[0])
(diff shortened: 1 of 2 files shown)


src/latexify/codegen/function_codegen.py
```diff
@@ -355,6 +355,9 @@ def visit_Call(self, node: ast.Call) -> str:
             (r"\mathrm{" + func_str + r"}\left(", r"\right)"),
         )
 
+        if func_str == "ndarray" or func_str == "array":
```
odashi, line 358:
```suggestion
        if func_str in ("ndarray", "array"):
```

src/latexify/codegen/function_codegen.py
```diff
@@ -363,6 +366,35 @@ def visit_Call(self, node: ast.Call) -> str:
         arg_strs = [self.visit(arg) for arg in node.args]
         return lstr + ", ".join(arg_strs) + rstr
 
+    def _generate_matrix(self, node: ast.Call) -> str:
```
odashi, line 369: `_generate_ndarray` makes sense more I think.

```suggestion
    def _generate_ndarray(self, node: ast.Call) -> str:
```

src/latexify/codegen/function_codegen.py
```diff
@@ -363,6 +366,35 @@ def visit_Call(self, node: ast.Call) -> str:
...
+        """Generates a matrix from numpy.matrix."""
```
odashi, line 370:
```suggestion
        """Generates a matrix from numpy.ndarray.
        
        Args:
            node: Call node representing a supported numpy function.
            
        Returns:
            Generated LaTeX expression.
        """
```

src/latexify/codegen/function_codegen.py
```diff
@@ -355,6 +355,9 @@ def visit_Call(self, node: ast.Call) -> str:
             (r"\mathrm{" + func_str + r"}\left(", r"\right)"),
         )
 
+        if func_str == "ndarray" or func_str == "array":
+            return self._generate_matrix(node)
```
odashi, line 359: I think we need to fall back to the default codegen if _generate_matric failed for some reason.
And it would be better to return None if _generate_matrix detected unsupported syntax, because raising the exception destroyes the entire process.
```suggestion
            latex = self._generate_matrix(node)
            if latex is not None:
                return latex
```

src/latexify/codegen/function_codegen.py
```diff
@@ -363,6 +366,35 @@ def visit_Call(self, node: ast.Call) -> str:
...
+    def _generate_matrix(self, node: ast.Call) -> str:
+        """Generates a matrix from numpy.matrix."""
+        if len(node.args) != 1:
+            # raise with function name
+            raise ValueError(
+                f"Invalid number of arguments {len(node.args)}"
+            )
+        
+        arg = node.args[0]
+        # TODO: Support string being passed to numpy.matrix.
```
odashi, line 378: Question: what is the "string" in this context?

src/latexify/codegen/function_codegen.py
```diff
@@ -363,6 +366,35 @@ def visit_Call(self, node: ast.Call) -> str:
...
+        if not isinstance(arg, ast.List):
+            raise exceptions.LatexifyNotSupportedError(
+                "Only matrices represented by lists are supported."
+            )
```
odashi, line 382: Place this block to the `else` clause of the following `if`

src/latexify/codegen/function_codegen.py
```diff
@@ -363,6 +366,35 @@ def visit_Call(self, node: ast.Call) -> str:
...
+        latex = r"\begin{bmatrix}"
+        if type(arg) == ast.List:
+            for elt in arg.elts:
+                if type(elt) == ast.List:
+                    for sub_elt in elt.elts:
+                        latex += self.visit(sub_elt) + r" & "
+                    latex = latex[:-2] + r" \\ "
+                else:
+                    latex += self.visit(elt) + r" & "
+            latex = latex[:-3] + r"\end{bmatrix}"
```
odashi, line 394: I think the following steps could be more intuitive and much efficient:

1. Checks if the arg is List, and all elts are List with the same length. Stops the process if the condition is not satisfied.
2. ```python
   # This should work without any breakages iff the condition 1 is satisfied.
   contents = r" \\ ".join(
       " & ".join(self.visit(x) for x in row.elts)
       for row in arg.elts
   )
   return r"\begin{bmatrix} " + contents + r" \end{bmatrix}"
   ```

odashi (commented):
Add sufficient unit tests in `function_codegen_test.py`.

src/latexify/codegen/function_codegen.py
kshxtij: [numpy.matrix ](https://numpy.org/doc/stable/reference/generated/numpy.matrix.html) supports passing in strings like 
and parsing that into a ndarray
```
a = np.matrix('1 2; 3 4')
```

src/latexify/codegen/function_codegen.py
kshxtij: I was thinking something like this to properly handle row, column and mxn matrices correctly. What do you think?

```python
        if isinstance(arg, ast.List):
            # handle row vectors
            if all(isinstance(x, ast.Constant) for x in arg.elts):
                contents = " & ".join([self.visit(x) for x in arg.elts])
                return r"\begin{bmatrix} " + contents + r" \end{bmatrix}"
            # handle column vector
            if all(isinstance(x, ast.List) for x in arg.elts) and all(len(x.elts) == 1 for x in arg.elts):
                contents = r" \\ ".join([self.visit(x.elts[0]) for x in arg.elts])
                return r"\begin{bmatrix} " + contents + r" \end{bmatrix}"
            # handle matrices
            elif all(isinstance(elt, ast.List) and len(elt.elts) == len(arg.elts) for elt in arg.elts):
                contents = r" \\ ".join(
                    " & ".join(self.visit(x) for x in row.elts)
                    for row in arg.elts
                )
                return r"\begin{bmatrix} " + contents + r" \end{bmatrix}"
        return None
```

Wubpooz (approved):

src/latexify/codegen/function_codegen.py
odashi: It looks still wrong I think:

- 1st condition looks too strict because it introduced another constraint that the member is only `Constant`.
- 2nd condition is not necessary if we handle $M \times N$ matrix.
- 3rd condition looks to handle only square matrices.

I tried to complete it, and noticed that it requrie more detailed checks. How about it:

無題

odashi: (btw, I think the above examples can be used in unit tests as well.)

odashi:
@kshxtij Hi, I'd appreciate it if you could proceed this pull request!

kshxtij:
I'll get this finished over the weekend, sorry it's been so delayed. Just very busy with deadlines and exams!

src/latexify/codegen/function_codegen.py
```diff
@@ -355,6 +355,30 @@ def visit_Call(self, node: ast.Call) -> str:
...
+
+            row0 = arg.elts[0]
+
+            if not isinstance(row0, ast.List):
+                return self._generate_ndarray([self.visit(x) for x in arg.elts])
+
+            if not row0.elts:
+                return None
+
+            nCols = len(row0.elts)
```
odashi, line 371: nCols is not allowed under the naming convention.
https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations
```suggestion
            ncols = len(row0.elts)
```

src/latexify/codegen/function_codegen.py
```diff
@@ -355,6 +355,30 @@ def visit_Call(self, node: ast.Call) -> str:
...
+
+            if not all(
+                isinstance(row, ast.List) and len(row.elts) == nCols for row in arg.elts
+            ):
+                return None
+
+            return self._generate_ndarray(
+                [[self.visit(x) for x in row.elts] for row in arg.elts]
+            )
```
odashi, line 380: This block must be placed on _generate_ndarray, and if it returns None we need to fall back to the default behavior.

odashi:
@kshxtij No problem! I just commented more. Could you fix CI errors as well?

odashi:
@kshxtij Also I'd appreciate it if you add every test case I commented above.

odashi:
Reminder: we will close this pull request tomorrow because of more than 1 week inactivity. If you are still working on this, please update the pull request.

odashi:
Closes this pull request, I'll take over the remaining work.

PR #120, opened 2022-11-18
Add isort module to checks script and Github CI

jianan1104:
# Overview

Add isort module to check python import order and format identically

# Details

- add isort module to development environment
- using isort to refactor the code
- add isort check to `checks.sh`
- add isort step to Github CI


odashi (approved):
Thanks!

PR #122, opened 2022-11-19
Show example in Readme

cool-RR:
Now when people visit this GitHub page, they'll immediately understand what this package does.


odashi (commented):
Thanks, this is definitely useful for users.

Though this is just a matter of taste, I prefer PNG a bit over JPEG for screenshots. I'd also be happy to merge it as-is if there's some intent to choose the current format.

cool-RR:
I tried converting it to png now, but I got some weird transperancy issues. I tried both Irfanview and Photoshop and got the same weird result. So I'd prefer that you merge it with jpg.

odashi (approved):
Great, sure I'd merge it as-is.

Issue #65, opened 2022-10-25
Support expanding some functions


odashi:
Follows #63 

It is good to provide a `expand_functions` boolean option to control whether some composite functions are expanded or not, e.g.:

- `hypot(x, y)` -> $\sqrt{x^2 + y^2}$
- `atan2(y, x)` -> $\arctan{\frac{y}{x}}$
- `expit(x)` -> $\frac{1}{1+\exp{(-x)}}$

MAKMED1337:
Is it better `expit(x)` to be $\frac{1}{1+e^{-x}}$ ?

odashi:
@MAKMED1337 Not sure, but it looks there should be another option to control the behavior around $\exp$ .

ZibingZhang:
List of methods in `math` that I think we should allow expansion for:

- `exp(x)` -> $e^x$
- `exp2(x)` -> $2^x$
- `expm1(x)` -> $\exp(x) - 1$
- `log1p(x)` -> $\log(1 + x)$
- `pow(x, y)` -> $x^y$
- `atan2(x)` -> $\arctan(y / x)$
- `erf` -> too complicated, also I don't think integrals are supported... yet
- `erfc` -> see above
- `gamma` -> maybe for positive integers? I think it's best to leave alone for now
- `lgamma` -> see above

Let me know what you think @odashi, and I'll get started on these

odashi:
I think non-elementary functions (`erf`, `erfc`, `gamma`, `lgamma`) should be avoided from this option. Other candidates looks fine to me.

odashi:
I also think our current implementation (#125) naturally solves https://github.com/google/latexify_py/issues/65#issuecomment-1290288048 , as users can choose whether `exp` is expanded or not.

odashi:
This is generally supported in the current implementation. We'd support individual functions in separate issues (if needed).

PR #125, opened 2022-11-19
Support expanding some functions

ZibingZhang:
# Overview

Adds a `NodeTransformer` that expands functions.

# Details

- Adds a `expand_functions` keyword argument onto `get_latex` which allows the user to specify the names of the methods they want to have expanded.
- Introduces a `class FunctionExpander(ast.NodeExpander)` which replaces `ast.Call` with the equivalent expanded `ast.AST`.
- Only expanded method to be supported is `hypot(x, y) -> \sqrt{x^2 + y^2}`.
- Supports nested `hypot` calls, e.g. `hypot(hypot(a, b), x, y)`.

### Thoughts

As discussed in the issue, there may be multiple ways to expand a function. We could expose various expansion methods, and accept them in the form `expanded_functions={"method1": Method1Expander1, "method2": None}` where `None` defaults to some predetermined expansion. This is a bit gross, but we could probably make this a bit nicer with the config refactor. This would also allow users to define their own expansion methods.

There isn't a way to prevent a stack overflow right now due to `f(x) -> g(x)` and `g(x) -> f(x)`. My initial thought would be to add a `function_expansion_depth=1` as a keyword argument, and it would be implemented like so:
- Initialize `current_depth` to 0 for each instance of `FunctionExpander`.
- In `visit_Call`:
  -  If `current_depth > function_expansion_depth`, then return early.
  - If any expansions are applied, increment `current_depth` by 1 (I'd err towards forcing users to specify if they want to try for deeper level of replacements).
    - After mapping over child nodes decrement `current_depth` by 1.

# References

https://github.com/google/latexify_py/issues/65

# Blocked by

Potentially blocked by https://github.com/google/latexify_py/pull/103. 

We could merge this in before, and then figure out a config format in the other PR, or vice versa.

### DIFF
--- a/src/integration_tests/regression_test.py
+++ b/src/integration_tests/regression_test.py
@@ -310,3 +310,23 @@ def test_expand_nested_function() -> None:
         r"x^{{2}} + y^{{2}}}"
     )
     _check_function(solve, latex, expand_functions={"hypot"})
+
+
+def test_docstring_allowed() -> None:
+    def solve(x):
+        """The identity function."""
+        return x
+
+    latex = r"\mathrm{solve}(x) = x"
+    _check_function(solve, latex)
+
+
+def test_multiple_constants_allowed() -> None:
+    def solve(x):
+        """The identity function."""
+        123
+        True
+        return x
+
+    latex = r"\mathrm{solve}(x) = x"
+    _check_function(solve, latex)
--- a/src/integration_tests/regression_test.py
+++ b/src/integration_tests/regression_test.py
@@ -7,7 +7,6 @@ from collections.abc import Callable
 from typing import Any
 
 from latexify import frontend
-from latexify.test_stubs import hypot
 
 
 def _check_function(
@@ -282,17 +281,27 @@ def test_sub_bracket() -> None:
     _check_function(solve, latex)
 
 
-def test_expand_hypot_function() -> None:
+def test_expand_hypot_function_without_attribute_access() -> None:
+    from math import hypot
+
     def solve(x, y, z):
         return hypot(x, y, z)
 
-    latex = r"\mathrm{solve}(x, y, z) = " r"\sqrt{x^{{2}} + y^{{2}} + z^{{2}}}"
-    _check_function(solve, latex, expand_functions=["hypot"])
+    latex = r"\mathrm{solve}(x, y, z) = \sqrt{x^{{2}} + y^{{2}} + z^{{2}}}"
+    _check_function(solve, latex, expand_functions={"hypot"})
+
+
+def test_expand_hypot_function() -> None:
+    def solve(x, y, z):
+        return math.hypot(x, y, z)
+
+    latex = r"\mathrm{solve}(x, y, z) = \sqrt{x^{{2}} + y^{{2}} + z^{{2}}}"
+    _check_function(solve, latex, expand_functions={"hypot"})
 
 
 def test_expand_nested_function() -> None:
     def solve(a, b, x, y):
-        return hypot(hypot(a, b), x, y)
+        return math.hypot(math.hypot(a, b), x, y)
 
     latex = (
         r"\mathrm{solve}(a, b, x, y) = "
@@ -300,4 +309,4 @@ def test_expand_nested_function() -> None:
         r"\sqrt{a^{{2}} + b^{{2}}}^{{2}} + "
         r"x^{{2}} + y^{{2}}}"
     )
-    _check_function(solve, latex, expand_functions=["hypot"])
+    _check_function(solve, latex, expand_functions={"hypot"})
--- a/src/integration_tests/regression_test.py
+++ b/src/integration_tests/regression_test.py
@@ -279,3 +279,24 @@ def test_sub_bracket() -> None:
         r"a - b \mathclose{}\right) - a b"
     )
     _check_function(solve, latex)
+
+
+def test_expand_hypot_function() -> None:
+    def solve(x, y, z):
+        return hypot(x, y, z)
+
+    latex = r"\mathrm{solve}(x, y, z) = " r"\sqrt{x^{{2}} + y^{{2}} + z^{{2}}}"
+    _check_function(solve, latex, expand_functions=["hypot"])
+
+
+def test_expand_nested_function() -> None:
+    def solve(a, b, x, y):
+        return hypot(hypot(a, b), x, y)
+
+    latex = (
+        r"\mathrm{solve}(a, b, x, y) = "
+        r"\sqrt{"
+        r"\sqrt{a^{{2}} + b^{{2}}}^{{2}} + "
+        r"x^{{2}} + y^{{2}}}"
+    )
+    _check_function(solve, latex, expand_functions=["hypot"])
--- a/src/integration_tests/regression_test.py
+++ b/src/integration_tests/regression_test.py
@@ -7,6 +7,7 @@ from collections.abc import Callable
 from typing import Any
 
 from latexify import frontend
+from latexify.test_stubs import hypot
 
 
 def _check_function(
--- a/src/latexify/ast_utils_test.py
+++ b/src/latexify/ast_utils_test.py
@@ -59,6 +59,37 @@ def test_make_constant_invalid() -> None:
         ast_utils.make_constant(object())
 
 
+@test_utils.require_at_most(7)
+@pytest.mark.parametrize(
+    "value,expected",
+    [
+        (ast.Bytes(s=b"foo"), True),
+        (ast.Constant("bar"), True),
+        (ast.Ellipsis(), True),
+        (ast.NameConstant(value=None), True),
+        (ast.Num(n=123), True),
+        (ast.Str(s="baz"), True),
+        (ast.Expr(value=ast.Num(456)), False),
+        (ast.Global("qux"), False),
+    ],
+)
+def test_is_constant_legacy(value: ast.AST, expected: bool) -> None:
+    assert ast_utils.is_constant(value) is expected
+
+
+@test_utils.require_at_least(8)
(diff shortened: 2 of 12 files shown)


src/integration_tests/regression_test.py
```diff
@@ -279,3 +280,24 @@ def solve(a, b):
...
+
+def test_expand_nested_function() -> None:
+    def solve(a, b, x, y):
+        return hypot(hypot(a, b), x, y)
+
+    latex = (
+        r"\mathrm{solve}(a, b, x, y) = "
+        r"\sqrt{"
+        r"\sqrt{a^{{2}} + b^{{2}}}^{{2}} + "
+        r"x^{{2}} + y^{{2}}}"
```
odashi, line 301: nit: Maybe it'd be good to apply some symbolic optimizations since nested `hypot` can be removed: $\sqrt{\sqrt{x^2 + y^2}^2 + z^2} = \sqrt{x^2 + y^2 + z^2}$.

This is beyond the scope of this pull request. We may need to make another issue to discuss about it.

src/latexify/codegen/latex_test.py
```diff
@@ -2,7 +2,8 @@
 
 from __future__ import annotations
 
-from latexify.codegen.latex import Latex  # Ignores [22-imports] for convenience.
+from latexify.codegen.latex import \
+    Latex  # Ignores [22-imports] for convenience.
```
odashi, line 6: some mistakes?
```suggestion
from latexify.codegen.latex import Latex  # Ignores [22-imports] for convenience.
```

src/latexify/test_stubs.py
```diff
@@ -0,0 +1,4 @@
+"""Stubbed methods for tests."""
```
odashi, line 1: Technically, we don't need this kind of file, because:

- Every function is evaluated only when it is invoked
- latexify does not evaluate any inner stuff of the given function: we only "evaluates" the source.

src/latexify/transformers/function_expander_test.py
```diff
@@ -0,0 +1,68 @@
+"""Tests for latexify.transformers.function_expander."""
+
+from __future__ import annotations
+
+import ast
+
+from latexify import parser, test_utils
+from latexify.test_stubs import hypot
```
odashi, line 8: If we need a definition of hypot for some reason (e.g., to pass linters), defining it directly in this file would be better.

src/latexify/transformers/function_expander_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+    expected = _make_ast(
+        ["x", "y"],
+        ast.Call(
+            ast.Name("sqrt"),
+            [
+                ast.BinOp(
+                    ast.BinOp(ast.Name("x"), ast.Pow(), ast.Num(2)),
+                    ast.Add(),
+                    ast.BinOp(ast.Name("y"), ast.Pow(), ast.Num(2)),
```
odashi, line 62: use `ast_utils.make_constant` instead. `Num` will be removed in the future version.
```suggestion
                    ast.BinOp(ast.Name("x"), ast.Pow(), ast_utils.make_constant(2)),
                    ast.Add(),
                    ast.BinOp(ast.Name("y"), ast.Pow(), ast_utils.make_constant(2)),
```

odashi (commented):
This is very good, and sorry for putting many comments (most of them are trivial I think)

src/latexify/transformers/function_expander.py
```diff
@@ -0,0 +1,66 @@
...
+    if len(node.args) == 0:
+        raise exceptions.LatexifyNotSupportedError(
+            "FunctionExpander does not support expanding 'hypot' with zero arguments."
+        )
+
+    args = []
+    for arg in node.args:
+        args.append(ast.BinOp(arg, ast.Pow(), ast.Num(2)))
+
+    args = list(map(lambda arg: function_expander.visit(arg), args))
```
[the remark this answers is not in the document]
ZibingZhang: very clean

src/integration_tests/regression_test.py
```diff
@@ -7,6 +7,7 @@
 from typing import Any
 
 from latexify import frontend
+from latexify.test_stubs import hypot
```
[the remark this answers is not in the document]
ZibingZhang: you're right, didn't realize that was a thing

src/latexify/codegen/latex_test.py
ZibingZhang: oh yeah black keeps on formatting it that way for some reason, I've had to `--assume-unchanged` but must've missed it here

src/latexify/test_stubs.py
ZibingZhang: I'll remove it for now since we have `math.hypot`, but I only added it because `pflake8` complains about it otherwise. It doesn't cause any runtime errors though, just linting ones.

odashi:
> There isn't a way to prevent a stack overflow right now due to f(x) -> g(x) and g(x) -> f(x)

Nice insight. For now we can introduce functions that don't employ recursive definitions, but we need to keep in mind that this could become a problem in the future. It is better to add some TODOs to `function_expander.py`.

src/integration_tests/regression_test.py
ZibingZhang: I would imagine that this would be a separate transformer that takes place before function expansion

odashi: Yeah, and it would become a more complex optimizer. We could borrow some knowledge from other symbolic math tools as well.

src/latexify/transformers/function_expander.py
```diff
@@ -0,0 +1,61 @@
...
+        return ast_utils.make_constant(0)
+
+    args = [
+        ast.BinOp(function_expander.visit(arg), ast.Pow(), ast_utils.make_constant(2))
+        for arg in node.args
+    ]
+
+    args_reduced = functools.reduce(lambda a, b: ast.BinOp(a, ast.Add(), b), args)
+    return ast.Call(
+        func=ast.Name(id="sqrt", ctx=Load()),
```
ZibingZhang, line 54: Maybe in `constants.py`, `BUILTIN_FUNCS` should be `dict[enum, tuple[str, str]]`? 

Basically what's happening here is that we're replacing `hypot` with `sqrt`, while also relying on the fact `sqrt` gets further preprocessed by getting replaced with `(r"\sqrt{", "}")`.

Having an enum rather than a bunch of strings might make the code a bit more readable, as well as helping to prevent errors with typos. In the future with more function expansions and different transformers, there might be a lot of cases where we care about specific, commonly-used functions, such as `sqrt`.

If you agree with changing it to an enum I'll tackle that in another issue / PR.

odashi: To be clear, do you mean defining an Enum class something like:
```python
class BuiltinFnNames(str, enum.Enum):
    ...
    SQRT = "sqrt"
```

and use it in:
```python
BUILTIN_FUNCS: dict[BuiltinFnNames, tuple[str, str]] = {
    ...
    BuiltinFnNames.SQRT: (r"\sqrt{", "}"),
    ...
}
```
? If I'm wrong please fix me.

ZibingZhang: something like that yeah

odashi: It looks nicer than the current impl. I guess we finally need to adopt more sophisticated way to represent functions, but for now the idea above looks a good choice.

src/latexify/ast_utils_test.py
```diff
@@ -146,3 +146,21 @@ def test_extract_int_invalid() -> None:
...
+    [
+        (ast.Call(ast.Name("hypot"), ast.Constant("foo")), "hypot"),
+        (
+            ast.Call(
+                ast.Attribute(ast.Name("math"), "hypot", ast.Load()),
+                ast.Constant("foo"),
+            ),
+            "hypot",
+        ),
+        (ast.Call(ast.Constant(123), ast.Constant("foo")), None),
```
odashi, line 162: - These `Call`s are illegal because the `args` parameter should take a list of `expr`.
- The `func` parameter could take any `expr`, but in this case the constant `123` is never populated here in the actual AST. I guess some other example is preferred, e.g., `ast.Call(func=ast.Name("foo"), args=[])`, which represents `foo()(...)`

odashi (commented):
Sorry for bothering again! Overall I like this change, but there are some comments before merging.

A general suggestion: if we instantiate AST nodes, it is preferred to explicitly specify parameter names for every argument. AST structure is sometimes updated, and without parameter names it may lack version compatibility (not only currently supported versions, but also versions in the future, in which we couldn't expect what kind of changes will be introduced)

ZibingZhang:
> Sorry for bothering again! Overall I like this change, but there are some comments before merging.

No worries at all! Thanks for taking the time to review :) I think this is a really cool project.

I'll take a look at this sometime later today / tomorrow. Do you have any recs for another issue for me to look at after this one? It seems like quite a few of them are blocked my issues that have existing PRs, but I'd love to work on something else if I've got free time this week

odashi:
@ZibingZhang I think this PR is almost independent from others except #103 which is an essential change. If you could take your time for that PR it's welcome.

odashi:
I also think adding supports for other functions in `math` and `numpy` can be processed independently (but blocked by this issue #125).

ZibingZhang:
> A general suggestion: if we instantiate AST nodes, it is preferred to explicitly specify parameter names for every argument.

Good point, will keep this in mind moving forward.

> I also think adding supports for other functions in math and numpy can be processed independently (but blocked by this issue https://github.com/google/latexify_py/pull/125).

That's a good point. I'll start working through that backlog.

odashi (approved):
Really good change, thanks!

Issue #104, opened 2022-11-14
Breaking on function with docstring


fccoelho:
I am getting the error below for the following function:
```python
@latexify.function(use_math_symbols=True)
def model(t: float, y: list, params: dict) -> list:
        """
        SIR Model.
        :param t: time step
        :param y: state of the model at time t
        :param params: parameter dictionary
        :return:
        """
        S, I, R = y
        beta, gamma, N = params['beta'], params['gamma'], params['N']
        return [
            -beta * S * I / N,
            beta * S * I / N - gamma * I,
            gamma * I
        ]
```

```python
LatexifyNotSupportedError: Codegen supports only Assign nodes in multiline functions, but got: Expr
```

If I remove the docstring, it works as expected

soosub:
I encountered the same issue and removing the docstring resolved it. Thanks for raising the issue @fccoelho! Indeed, it would be nice to keep docstrings *and* latexify functionality all at once😌

PS cheers for maintaining this package @odashi et al., it's awesome!

Casper-Guo:
cc:@lakeblair 

Maybe you can investigate this issue

LakeBlair:
@odashi Hi! I can take a look at this issue and see if I can find a solution to it.

odashi:
Hi @LakeBlair , feel free to propose a pull request for this issue.

odashi:
Resolved by #126.

PR #126, opened 2022-11-20
Allow docstring in function definition

ZibingZhang:
# Overview

Allows a docstring to be defined at the top of the function.

# Details

Docstrings (multiline strings) are represented as an `ast.Expr(ast.Constant)`.
When validating that all child statements (minus the last one) in a function definition are of the class `ast.Assign`, we make an exception if the first one is an `ast.Expr(ast.Constant)`.

# References

https://github.com/google/latexify_py/issues/104

### DIFF
--- a/src/latexify/ast_utils.py
+++ b/src/latexify/ast_utils.py
@@ -41,6 +41,31 @@ def make_constant(value: Any) -> ast.expr:
     raise ValueError(f"Unsupported type to generate Constant: {type(value).__name__}")
 
 
+def is_constant(node: ast.AST) -> bool:
+    """Checks if the node is a constant.
+
+    Args:
+        node: The node to examine.
+
+    Returns:
+        True if the node is a constant, False otherwise.
+    """
+    if sys.version_info.minor < 8:
+        if isinstance(
+            node,
+            (ast.Bytes, ast.Constant, ast.Ellipsis, ast.NameConstant, ast.Num, ast.Str),
+        ):
+            return True
+    else:
+        if isinstance(node, ast.Constant):
+            return True
+
+    if isinstance(node, ast.Expr):
+        return is_constant(node.value)
+
+    return False
+
+
 def extract_int_or_none(node: ast.expr) -> int | None:
     """Extracts int constant from the given Constant node.
 
--- a/src/latexify/ast_utils.py
+++ b/src/latexify/ast_utils.py
@@ -51,19 +51,12 @@ def is_constant(node: ast.AST) -> bool:
         True if the node is a constant, False otherwise.
     """
     if sys.version_info.minor < 8:
-        if isinstance(
+        return isinstance(
             node,
             (ast.Bytes, ast.Constant, ast.Ellipsis, ast.NameConstant, ast.Num, ast.Str),
-        ):
-            return True
+        )
     else:
-        if isinstance(node, ast.Constant):
-            return True
-
-    if isinstance(node, ast.Expr):
-        return is_constant(node.value)
-
-    return False
+        return isinstance(node, ast.Constant)
 
 
 def extract_int_or_none(node: ast.expr) -> int | None:
--- a/src/latexify/ast_utils_test.py
+++ b/src/latexify/ast_utils_test.py
@@ -69,11 +69,11 @@ def test_make_constant_invalid() -> None:
         (ast.NameConstant(value=None), True),
         (ast.Num(n=123), True),
         (ast.Str(s="baz"), True),
-        (ast.Expr(value=ast.Num(456)), True),
+        (ast.Expr(value=ast.Num(456)), False),
         (ast.Global("qux"), False),
     ],
 )
-def test_is_constant_legacy(value: Any, expected: ast.Constant) -> None:
+def test_is_constant_legacy(value: ast.AST, expected: bool) -> None:
     assert ast_utils.is_constant(value) is expected
 
 
@@ -82,11 +82,11 @@ def test_is_constant_legacy(value: Any, expected: ast.Constant) -> None:
     "value,expected",
     [
         (ast.Constant("foo"), True),
-        (ast.Expr(value=ast.Constant(123)), True),
+        (ast.Expr(value=ast.Constant(123)), False),
         (ast.Global("bar"), False),
     ],
 )
-def test_is_constant(value: Any, expected: ast.Constant) -> None:
+def test_is_constant(value: ast.AST, expected: bool) -> None:
     assert ast_utils.is_constant(value) is expected
 
 
--- a/src/latexify/ast_utils_test.py
+++ b/src/latexify/ast_utils_test.py
@@ -59,6 +59,37 @@ def test_make_constant_invalid() -> None:
         ast_utils.make_constant(object())
 
 
+@test_utils.require_at_most(7)
+@pytest.mark.parametrize(
+    "value,expected",
+    [
+        (ast.Bytes(s=b"foo"), True),
+        (ast.Constant("bar"), True),
+        (ast.Ellipsis(), True),
+        (ast.NameConstant(value=None), True),
+        (ast.Num(n=123), True),
+        (ast.Str(s="baz"), True),
+        (ast.Expr(value=ast.Num(456)), True),
+        (ast.Global("qux"), False),
+    ],
+)
+def test_is_constant_legacy(value: Any, expected: ast.Constant) -> None:
+    assert ast_utils.is_constant(value) is expected
+
+
+@test_utils.require_at_least(8)
+@pytest.mark.parametrize(
+    "value,expected",
+    [
+        (ast.Constant("foo"), True),
+        (ast.Expr(value=ast.Constant(123)), True),
+        (ast.Global("bar"), False),
+    ],
+)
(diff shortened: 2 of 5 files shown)


src/latexify/codegen/function_codegen.py
```diff
@@ -252,7 +252,15 @@ def visit_FunctionDef(self, node: ast.FunctionDef) -> str:
         body_strs: list[str] = []
 
         # Assignment statements (if any): x = ...
-        for child in node.body[:-1]:
+        for i, child in enumerate(node.body[:-1]):
+            # Allow docstrings
+            if (
+                i == 0
```
odashi, line 258: I think this condition is not necessary. In general all `Expr` with `Constant` has no side effects and can be skipped.

src/latexify/codegen/function_codegen_test.py
```diff
@@ -40,6 +40,25 @@ def test_visit_functiondef_use_signature() -> None:
...
+    tree = ast.FunctionDef(
+        name="f",
+        args=ast.arguments(
+            args=[ast.arg(arg="x")],
+            kwonlyargs=[],
+            kw_defaults=[],
+            defaults=[],
+        ),
+        body=[
+            ast.Expr(ast_utils.make_constant("""docstring""")),
```
odashi, line 53:
```suggestion
            ast.Expr(ast_utils.make_constant("docstring")),
```

src/latexify/codegen/function_codegen_test.py
```diff
@@ -40,6 +40,25 @@ def test_visit_functiondef_use_signature() -> None:
...
+            ast.Return(value=ast.Name(id="x", ctx=ast.Load())),
+        ],
+        decorator_list=[],
+    )
+    latex_with_flag = r"\mathrm{f}(x) = x"
+    assert FunctionCodegen().visit(tree) == latex_with_flag
```
odashi, line 59: We don't have any "flag" here.

```suggestion
    latex = r"\mathrm{f}(x) = x"
    assert FunctionCodegen().visit(tree) == latex
```

src/latexify/codegen/function_codegen_test.py
```diff
@@ -40,6 +40,25 @@ def test_visit_functiondef_use_signature() -> None:
     assert FunctionCodegen(use_signature=True).visit(tree) == latex_with_flag
 
 
+def test_visit_functiondef_ignore_docstring() -> None:
```
odashi, line 43: Add another test to check if the `Constant` in the middle of function can be skipped as well.

odashi (commented):
It looks basically good!

src/latexify/codegen/function_codegen.py
```diff
@@ -252,7 +252,15 @@ def visit_FunctionDef(self, node: ast.FunctionDef) -> str:
...
+                and isinstance(child, ast.Expr)
+                and isinstance(child.value, ast.Constant)
```
odashi, line 260: `Constant` is not available in 3.7 (I noticed it through the CI failure).

I think it'd be good if we have a function, e.g., `ast_utils.is_constant(node)` to check if the node is Constant or not.

src/latexify/codegen/function_codegen_test.py
```diff
@@ -40,6 +40,25 @@ def test_visit_functiondef_use_signature() -> None:
...
+    tree = ast.FunctionDef(
+        name="f",
+        args=ast.arguments(
+            args=[ast.arg(arg="x")],
+            kwonlyargs=[],
```
ZibingZhang, line 48: Thoughts on removing these empty fields from the tests? Would make it easier to read imo / remove some bloat

src/latexify/codegen/function_codegen.py
ZibingZhang: Oversight on my part. It's [available](https://github.com/python/cpython/blob/v3.7.15/Python/Python-ast.c#L962), but [`ast.Str`](https://github.com/python/cpython/blob/v3.7.15/Python/Python-ast.c#L948) isn't a `ast.Constant` for some reason.

odashi: Oh yeah, both I and you are partially correct. Some node classes are integrated into `Constant` in 3.8. In the official document:

> Deprecated since version 3.8: Old classes ast.Num, ast.Str, ast.Bytes, ast.NameConstant and ast.Ellipsis are still available, but they will be removed in future Python releases. In the meantime, instantiating them will return an instance of a different class.

https://docs.python.org/3.10/library/ast.html#node-classes

src/latexify/codegen/function_codegen_test.py
odashi: How about using `ast.parse` as in some other tests?

```python
tree = ast.parse(textwrap.dedent("""\
    def f(x):
        '''docstring'''
        return x
    """)
```

src/latexify/ast_utils.py
```diff
@@ -41,6 +41,31 @@ def make_constant(value: Any) -> ast.expr:
...
+            (ast.Bytes, ast.Constant, ast.Ellipsis, ast.NameConstant, ast.Num, ast.Str),
+        ):
+            return True
+    else:
+        if isinstance(node, ast.Constant):
+            return True
+
+    if isinstance(node, ast.Expr):
+        return is_constant(node.value)
+
```
odashi, line 65: This check is too much. This function should check only if the given node is the Constant or not to avoid unintended behavior.

src/latexify/ast_utils.py
```diff
@@ -41,6 +41,31 @@ def make_constant(value: Any) -> ast.expr:
...
+    Returns:
+        True if the node is a constant, False otherwise.
+    """
+    if sys.version_info.minor < 8:
+        if isinstance(
+            node,
+            (ast.Bytes, ast.Constant, ast.Ellipsis, ast.NameConstant, ast.Num, ast.Str),
+        ):
+            return True
```
odashi, line 58:
```suggestion
        return isinstance(
            node,
            (ast.Bytes, ast.Constant, ast.Ellipsis, ast.NameConstant, ast.Num, ast.Str),
        )
```

src/latexify/ast_utils.py
```diff
@@ -41,6 +41,31 @@ def make_constant(value: Any) -> ast.expr:
...
+    else:
+        if isinstance(node, ast.Constant):
+            return True
```
odashi, line 61:
```suggestion
        return isinstance(node, ast.Constant)
```

src/latexify/codegen/function_codegen_test.py
ZibingZhang:
```python
def test() -> None:
    tree = ast.parse(
        """
def f(x):
    '''docstring'''
    return x"""
    )
```

I've gone with this format. Unfortunately due to the nature of multiline strings it seems like the `def` can't have any preceding indentation, which makes it a bit awkward to read, but still better than before.

There can't be any space between the last non-whitespace character and the closing `"""`.

I decided to opt against the starting `\` because it doesn't actually add anything in terms of the test, and I felt that it's better to keep it simple for readability.

src/latexify/ast_utils.py
```diff
@@ -41,6 +41,31 @@ def make_constant(value: Any) -> ast.expr:
...
+
+    if isinstance(node, ast.Expr):
+        return is_constant(node.value)
+
+    return False
```
odashi, line 66: This statement is not necessary if we applied all suggestions above.

src/latexify/ast_utils_test.py
```diff
@@ -59,6 +59,37 @@ def test_make_constant_invalid() -> None:
...
+        (ast.Constant("bar"), True),
+        (ast.Ellipsis(), True),
+        (ast.NameConstant(value=None), True),
+        (ast.Num(n=123), True),
+        (ast.Str(s="baz"), True),
+        (ast.Expr(value=ast.Num(456)), True),
+        (ast.Global("qux"), False),
+    ],
+)
+def test_is_constant_legacy(value: Any, expected: ast.Constant) -> None:
```
odashi, line 76:
```suggestion
def test_is_constant_legacy(value: ast.AST, expected: bool) -> None:
```

src/latexify/ast_utils_test.py
```diff
@@ -59,6 +59,37 @@ def test_make_constant_invalid() -> None:
...
+@test_utils.require_at_least(8)
+@pytest.mark.parametrize(
+    "value,expected",
+    [
+        (ast.Constant("foo"), True),
+        (ast.Expr(value=ast.Constant(123)), True),
+        (ast.Global("bar"), False),
+    ],
+)
+def test_is_constant(value: Any, expected: ast.Constant) -> None:
```
odashi, line 89:
```suggestion
def test_is_constant(value: ast.AST, expected: bool) -> None:
```

src/latexify/codegen/function_codegen_test.py
odashi: > could probably add this function into test_utils if it's commonly used

I think this is not recommended according to the DAMP principle. It is somewhat desirable to write self-contained tests:
https://abseil.io/resources/swe-book/html/ch12.html#tests_and_code_sharing_dampcomma_not_dr

odashi (approved):
Great, thanks!

PR #128, opened 2022-11-21
docs(README.md): add nice badges to README

jianan1104:
# Overview

add nice badges to README

# Details
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # latexify
-[![Python](https://img.shields.io/pypi/pyversions/latexify-py.svg?style=plastic)](https://badge.fury.io/py/latexify-py)
+[![Python](https://img.shields.io/pypi/pyversions/latexify-py.svg)](https://pypi.org/project/latexify-py/)
 [![PyPI Latest Release](https://img.shields.io/pypi/v/latexify-py.svg)](https://pypi.org/project/latexify-py/)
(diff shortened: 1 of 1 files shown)


odashi:
I'm a bit thinking if it's better that the badges are placed on the header (before the title) or not:

無題

jianan1104:
> I'm a bit thinking if it's better that the badges are placed on the header (before the title) or not. WDYT
> 
> 無題

I think it's okay for any style. most github repos are let badges under the title. btw, I think we need a pretty avatar or cover
image

odashi:
Great, I have no strong opinions and it'd be okay to go with the most common way.

README.md
```diff
@@ -1,4 +1,11 @@
-# `latexify`
+# latexify
+[![Python](https://img.shields.io/pypi/pyversions/latexify-py.svg?style=plastic)](https://badge.fury.io/py/latexify-py)
```
odashi, line 2: The fury.io address actually points to the PyPI. We don't need to add additional third-party address.

```suggestion
[![Python](https://img.shields.io/pypi/pyversions/latexify-py.svg)](https://pypi.org/project/latexify-py/)
```

jianan1104:
> Great, I have no strong opinions and it'd be okay to go with the most common way.

Sure, let's do it.

odashi (approved):

PR #130, opened 2022-11-21
Add BuiltinFnEnum

ZibingZhang:
# Overview

Creates an enum class `BuiltinFnEnum` which lets us easily reuse common function names.

# Details

We can see this already being used in the function expander like so:

```python
return ast.Call(
    func=ast.Name(id=constants.BuiltinFnName.SQRT.value, ctx=ast.Load()),
    args=[args_reduced],
)
```

As more function expanders are written, this will be used more.

# References

Inspired from this conversation: https://github.com/google/latexify_py/pull/125#discussion_r1027275178

# Blocked by

None

### DIFF
--- a/src/latexify/constants.py
+++ b/src/latexify/constants.py
@@ -2,39 +2,79 @@
 
 from __future__ import annotations
 
+import enum
+
 PREFIXES = ["math", "numpy", "np"]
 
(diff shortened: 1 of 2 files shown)


odashi (approved):
Thanks!

PR #132, opened 2022-11-22
Expand elementary functions in math module

ZibingZhang:
__feature is completed, just needs more tests__

# Overview

Adds more functions which can be expanded.

# Details

# References

https://github.com/google/latexify_py/issues/65

# Blocked by

None

### DIFF
--- a/src/latexify/transformers/function_expander.py
+++ b/src/latexify/transformers/function_expander.py
@@ -7,7 +7,7 @@ from collections.abc import Callable
 from latexify import ast_utils, constants, exceptions
 
 
-# TODO(ZibingZhang): handle recursive function expansions
+# TODO(ZibingZhang): handle mutually recursive function expansions
 class FunctionExpander(ast.NodeTransformer):
     """NodeTransformer to expand functions.
 
--- a/src/latexify/transformers/function_expander.py
+++ b/src/latexify/transformers/function_expander.py
@@ -45,13 +45,16 @@ def _atan2_expander(function_expander: FunctionExpander, node: ast.Call) -> ast.
             "FunctionExpander only supports expanding 'atan2' with two arguments"
         )
 
+    print(node.args[0])
     return ast.Call(
         func=ast.Name(id=constants.BuiltinFnName.ATAN.value, ctx=ast.Load()),
         args=[
-            ast.BinOp(
-                left=function_expander.visit(node.args[0]),
-                op=ast.Div(),
-                right=function_expander.visit(node.args[1]),
+            function_expander.visit(
+                ast.BinOp(
+                    left=node.args[0],
+                    op=ast.Div(),
+                    right=node.args[1],
+                )
             )
         ],
     )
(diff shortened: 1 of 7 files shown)


ZibingZhang:
Not sure how much test coverage is best, but I've decided it might be easiest to test the base expansion in `function_expander_test` (as well as in `regression_test`), and that testing that nested functions get expanded properly should be done in `regression_test`. This is so we don't have to deal with constructing `ast`s by hand.

Not sure what the best way to test that for all these functions you need exactly `n` arguments. The last PR used the test method `test_hypot_expanded_no_args`, but it'd be nice if there was a way to specify any number of args except for `n` would raise an error.

Also maybe it would be nice to include an `ast_utils.make_name`, that way we can stop setting `ctx=ast.Load()` for every single one of them.

src/latexify/transformers/function_expander.py
```diff
@@ -39,6 +39,71 @@ def visit_Call(self, node: ast.Call) -> ast.AST:
         return node
 
 
+def _atan2_expander(function_expander: FunctionExpander, node: ast.Call) -> ast.AST:
+    if len(node.args) != 2:
+        raise exceptions.LatexifyNotSupportedError(
+            "FunctionExpander only supports expanding 'atan2' with two arguments"
+        )
```
odashi, line 46: Should raise `LatexifySyntaxError` because `atan2` takes exactly 2 arguments.
```suggestion
        raise exceptions.LatexifySyntaxError(
            "Incorrect number of arguments for atan2."
        )
```

Ditto for other functions.

I also guess it is better to introduce a validator function to check if the number of arguments is correct or not:

```python
def check_num_args(node: ast.Call, nargs: int) -> None:
    if len(node.args) != nargs:
        fn_name = ast_utils.extract_function_name_or_none(node)
        raise exceptions.LatexifySyntaxError(
            f"Incorrect number of arguments for {fn_name}."
            f" expected: {nargs}, but got {len(node.args)}"
```

src/latexify/transformers/function_expander.py
```diff
@@ -39,6 +39,71 @@ def visit_Call(self, node: ast.Call) -> ast.AST:
...
+
+    print(node.args[0])
```
odashi, line 48:
```suggestion
```

src/latexify/transformers/function_expander.py
```diff
@@ -39,6 +39,71 @@ def visit_Call(self, node: ast.Call) -> ast.AST:
...
+    return ast.Call(
+        func=ast.Name(id=constants.BuiltinFnName.ATAN.value, ctx=ast.Load()),
+        args=[
+            function_expander.visit(
+                ast.BinOp(
+                    left=node.args[0],
+                    op=ast.Div(),
+                    right=node.args[1],
+                )
+            )
```
odashi, line 58:
```suggestion
            ast.BinOp(
                left=function_expander.visit(node.args[0]),
                op=ast.Div(),
                right=function_expander.visit(node.args[1]),
            )
```

src/latexify/transformers/function_expander_test.py
```diff
@@ -106,3 +181,24 @@ def f():
...
+                ast.BinOp(
+                    left=ast_utils.make_constant(1),
+                    op=ast.Add(),
+                    right=ast.Name(id="x", ctx=ast.Load()),
+                )
+            ],
+        ),
+    )
+    transformed = FunctionExpander({"log1p"}).visit(parser.parse_function(f))
+    test_utils.assert_ast_equal(transformed, expected)
```
odashi, line 204: No unit test for `pow`.

odashi (approved):
Great, thanks!

PR #133, opened 2022-11-23
Move `expand_functions` to `Config`

odashi:
# Overview

Follow-up of #103 and #125: Adds `expand_functions` to `Config`.


ZibingZhang (approved):
maybe out of the scope of this PR, but might be nice to allow users to somehow specify custom function expansions

odashi:
@ZibingZhang As you said it is off-topic here, could you make another issue for it?

ShigekiKarita (approved):

PR #136, opened 2022-11-26
Add support for \infty

glebodin:
# Overview
Add Infinity as latex \infty to math constants and fix noninclusive range to \infty to be without -1.

# References

# Blocked by

### DIFF
--- a/src/latexify/codegen/function_codegen.py
+++ b/src/latexify/codegen/function_codegen.py
@@ -621,8 +621,8 @@ class FunctionCodegen(ast.NodeVisitor):
         if range_info.stop_int is None:
             stop = range_info.stop
             # use special processing if stop is infinity
-            if isinstance(range_info.stop, ast.Name) and str(stop.id) == "infty":
-                upper = "{" + self.visit(stop) + "}"
+            if isinstance(range_info.stop, ast.Name) and self.visit_Name(stop) == "{\\infty}":
+                upper = self.visit(stop)
             # use special processing if stop involves addition or subtraction
             elif isinstance(stop, ast.BinOp) and isinstance(
                 stop.op, (ast.Add, ast.Sub)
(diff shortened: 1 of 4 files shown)


src/latexify/codegen/function_codegen.py
```diff
@@ -619,13 +619,17 @@ def _get_sum_prod_range(self, node: ast.comprehension) -> tuple[str, str] | None
             lower_rhs = f"{{{range_info.start_int}}}"
 
         if range_info.stop_int is None:
-            # use special processing if range_info.stop involves addition or subtraction
-            if isinstance(range_info.stop, ast.BinOp) and isinstance(
-                range_info.stop.op, (ast.Add, ast.Sub)
+            stop = range_info.stop
+            # use special processing if stop is infinity
+            if isinstance(range_info.stop, ast.Name) and str(stop.id) == "infty":
+                upper = "{" + self.visit(stop) + "}"
```
odashi, line 625: This is always applied regardless of the `use_math_symbols` option. It would be better to maintain the original behavior if the flag is off.

src/latexify/codegen/function_codegen_test.py
```diff
@@ -197,6 +197,10 @@ def test_visit_setcomp(code: str, latex: str) -> None:
             "(i for i in range(n))",
             r"_{i = {0}}^{{n - 1}} \mathopen{}\left({i}\mathclose{}\right)",
         ),
+        (
+            "(i for i in range(infty))",
+            r"_{i = {0}}^{{infty}} \mathopen{}\left({i}\mathclose{}\right)",
```
odashi, line 202: Maybe this case fails.

odashi (commented):
Thanks!

src/latexify/codegen/function_codegen.py
glebodin: fixed

src/latexify/codegen/function_codegen_test.py
glebodin: fixed

src/latexify/codegen/function_codegen.py
odashi: It is recommended to check the flag directly rather than checking the result of visit since visit is not stable (as it is affected by any flags).

jonbarron:
Any reason why this PR never landed? I just encountered this issue myself, and having support for infty would be very nice.

PR #139, opened 2022-11-28
IdentifierConverter

### LINKED ISSUE #82: Fine-grained name processing
There are still some issues around name processing:

In the current implementation, only the function names are processed separately, but I think that that behavior is generally preferred for all identifiers, and it should be applied *by default*.

In some specific cases, it would be still good to process underscores (e.g., trailing numbers: `x_123` -> $x_{123}$), but rather this feature should be the option than retaining underscores.

### LINKED ISSUE #89: Wrong behavior around multiplication
The current implementation converts `a * b` into $ab$. This only works if the variables have only 1 character and doesn't generate a correct syntax for other cases:

* Long names: `abc * def` --> $abcdef$, which is hard to distinguish from other combination of variables.
* Unary operators: `x * -y` --> $x - y$, which is definitely wrong.

These cases require explicit multiply operator ( $\cdot$ or $\times$ ) between the operands.

I think the `Mult(left, right)` subtree should be converted to $left \cdot right$ *by default*, and try to avoid the operator only in the following cases:

- The left-hand side operand is a single character or a bracket (with or without unary operators and/or super/subscripts)
- The right-hand side operand is a single character or a bracket (with or without super/subscripts, *without* unary operators)

This requires fine-graind name processing, and may be applied after #82.

odashi:
# Overview

Introduces a common processing for all identifiers.

# Details

This change introduces `IdentifierConverter`, which converts identifier strings into:
- `foo` --> `\foo`, if `use_math_symbols == True` and the given identifier matches a supported math symbol name.
- `x` --> `x`, if the given identifier is exactly 1 character (except `_`)
- `foo_bar` --> `\mathrm{foo\_bar}`, otherwise.

This change also removes `use_raw_function_name` flag: the rule is applied to all identifiers (not only function names) without exceptions.

`IdentifierConverter` also returns a boolean flag to express whether the returned expression can be treated as a single character or not. This is potentially used to fix #89 .

# References

- Fixes #82

# Blocked by

- NA


ShigekiKarita (approved):
FYI sympy seems to use \text instead of \mathrm for identifiers (but not always :thinking:) https://github.com/sympy/sympy/blob/8ad4f490fecc269eac69a6976b50f44396f6a1ad/sympy/printing/tests/test_latex.py#L2203
https://github.com/sympy/sympy/issues/15934

odashi:
@ShigekiKarita I think `\text` is used for rendering regular texts with the default font of texts (not constraining it to Roman). In our case, I think always using `\mathrm` would be better, though I'm not sure how they chose both commands.

PR #143, opened 2022-12-03
NDArray support

### LINKED ISSUE #78: NDArray support

odashi:
# Overview

Supports following syntax (for example):

# Details

- Adds special treatment for `array` and `ndarray` functions to generate matrix expression.

# References

- Taken over from #118, thanks for initial implementation!
- Fixes #78 

# Blocked by

- NA


ShigekiKarita (approved):
VG

PR #144, opened 2022-12-03
Adds parse_expr utility

odashi:
# Overview

Adds `ast_utils.parse_expr`, which parses an expression and generates a corresponding AST.
This function prevents writing a tricky subscription `ast.parse(code).body[0].value` in some tests.

# Details

NA

# References

NA

# Blocked by

- #143


ShigekiKarita (approved):
LGTM

Issue #93, opened 2022-11-12
Support Match


odashi:
It would be great if we support the `match` clause added in Python 3.10, but the entire syntax is somewhat complex.

https://docs.python.org/3/library/ast.html#pattern-matching

erica-w-fu:
Could I get assigned to this issue? I would love to contribute to this repo!

odashi:
@erica-w-fu Hi, feel free to work on this!
Since the overall syntax around Match is very huge, I'd like to decompose it to small developments (pull requests) mainly for ease of reviewing.

erica-w-fu:
Could I get some clarification on what is expected? Are we trying to convert match statements to latex? What does support mean exactly?

odashi:
Basically implementing some codegen rules in `function_codegen.py` is all you need.

As for "small development", you can start to implement only a specific portion in the whole syntax (e.g., I think supporting only [MatchValue](https://docs.python.org/3/library/ast.html#ast.MatchValue) is the easiest one).
For all other cases, we should raise `LatexifyNotSupportedError`, which marks "not supported, but we'll support it in the future". For example:

```python
class FunctionCodegen(NodeVisitor):
    ...
    def visit_Match(self, node: ast.Match) -> str:
        if isinstance(node.pattern, ast.MatchValue):
            ...something...
        else:
            raise exceptions.LatexifyNotSupportedError("Unsupported match syntax.")
```

odashi:
I think the whole match case can't be supported in `latexify.function` because there are capture clauses that there aren't any corresponding syntax in math. It would be better to support only syntaxes with:
- Constants
- sequence/dict with only constants

odashi:
The resulting LaTeX could be expressed semilarly to that of the if clause:

```python
match x:
    case 0:
        return 1
    case _:
        return math.sin(x)/x
```

$$\left\\{ \begin{array}{ll} x, & \mathrm{if} \ x = 0 \\\\ \frac{\sin(x)}{x}, & \mathrm{otherwise} \end{array} \right.$$

erica-w-fu:
How can we build the project for development? Are there any detailed docs that explain how to build and test latexify/our implementation?

odashi:
@erica-w-fu I will add some documentation.

erica-w-fu:
Thanks! I was able to run all of the tests. Are there any build tools and commands that you use to test within jupyter notebook?

erica-w-fu:
[...]

PR #146, opened 2022-12-04
Match, match_case, and MatchValue

erica-w-fu:
Match, match_case, and MatchValue are implemented and tested.

# Overview

All of these produce the correct latex code when a user enters match python code with cases that are only matching the values.

# Details

Added three new functions to function_codegen.py. 
visit_Match provides the overall structure for the case statements in latex and loops through different match cases
visit_match_case separates the condition and return value and calls whichever type of match is being called
visit_MatchValue compares the value of the condition

Added two tests to function_codegen_test.py
test_matchvalue(), tests if one MatchValue case will return the correct latex code
test_multiple_matchvalue(), tests if multiple MatchValue cases will return the correct latex code

We worked with four collaborators 
@erica-w-fu
@lucybean-hi
@juliawgraham
@Yuqi07

# References

#93


Casper-Guo:
Out of curiosity, is this for 15-313?

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,30 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
...
+        for match_case in node.cases:
+            true_latex, cond_latex = self.visit(match_case)
+            if cond_latex:
+                latex += true_latex + r", & \mathrm{if} \ " + subject_latex + cond_latex + r" \\ "
+            else:
+                latex += true_latex + r", & \mathrm{otherwise}"
+        latex += r"\end{array} \right."
+        return latex
+    
+    def visit_match_case(self, node: ast.match_case) -> str:
```
odashi, line 552: Wrong type: this function returns `tuple[str, str]`. If you need to implement a method that returns other types than `str`, please implement non-visitor function (functions that don't have `visit_` prefix).

I also think this visitor is not necessary. Since the `cases` member is always `match_case` and `Match` is the only syntax where `match_case` is used, it could be implemented directly in `visit_Match`.

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,30 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
         latex += self.visit(node)
         return latex + r", & \mathrm{otherwise} \end{array} \right."
 
+    def visit_Match(self, node: ast.Match) -> str:
+        """Visit a match node"""
+        latex = r"\left\{ \begin{array}{ll} "
+        subject_latex = self.visit(node.subject)        
+        for match_case in node.cases:
```
odashi, line 543: In the current implementation of this library, we don't allow fall-through syntax, and the statement must return any value. In the case of match, it should have either:

- Cases that cover every possible conditions
- Cases with wildcards that are converted to `otherwise`

This is necessary to express the correct math expressions, and we need to check if the given code is correctly expressing this syntax.

Since the first option is hard to check, the second option is the only approach we could have: deny all `Match` subtrees without wildcards.

src/latexify/codegen/function_codegen_test.py
```diff
@@ -744,3 +743,35 @@ def test_use_set_symbols_compare(code: str, latex: str) -> None:
...
     assert function_codegen.FunctionCodegen(use_set_symbols=True).visit(tree) == latex
+
+
+def test_matchvalue() -> None:
+    tree = ast.parse(
+        textwrap.dedent(
+        """
+        match x:
+            case 0:
+                return 1
```
odashi, line 754: So this (and the following syntax) should raise the error, according to the comment above.

odashi (commented):
Thanks! Overall the policy looks fine, but there is some essential points to fix. Please take a look.

juliawgraham:
We have updated the code according to your comments. More specifically, on this PR, we merged visit_match_case into visit_Match. On PR #147 , we added error handling for wildcards. More specifically, wildcards could only appear, and must appear, as the last match case. We also updated our unit tests to reflect these changes.

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,23 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
         latex += self.visit(node)
         return latex + r", & \mathrm{otherwise} \end{array} \right."
 
+    def visit_Match(self, node: ast.Match) -> str:
+        """Visit a match node"""
+        latex = r"\left\{ \begin{array}{ll} "
+        subject_latex = self.visit(node.subject)        
+        for i, match_case in enumerate(node.cases):
```
odashi, line 543: the enumerator `i` is never used.

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,23 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
...
+            true_latex = self.visit(match_case.body[0])
```
odashi, line 544: we need to check if `match_case.body` has exactly one statement, or raise an appropriate exception in `latexify.exceptions`.
we also need to add a test case to capture this workflow.

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,23 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
...
+            cond_latex = self.visit(match_case.pattern)
+            latex += true_latex + r", & \mathrm{if} \ " + subject_latex + cond_latex + r" \\ "
+            
+        latex += r"\end{array} \right."
+        return latex
+
+    def visit_MatchValue(self, node: ast.MatchValue) -> str:
+        """Visit a MatchValue node"""
+        latex = self.visit(node.value)
+        return r" = " + latex
```
odashi, line 554:
```suggestion
        return " = " + latex
```

odashi:
Could you merge the main to resolve conflicts as well?

juliawgraham:
We have updated the code according to your comments and resolved the merge conflicts shown.

odashi:
And resolve all CI errors.

odashi (approved):
Note: This pull request involves errors, we need to merge #147 ASAP

odashi:
Will fix test errors by myself.

PR #147, opened 2022-12-04
MatchAs

erica-w-fu:
MatchAs is implemented and tested

# Overview

All of these produce the correct latex code when a user enters a wildcard case in a python match statement. A non-empty MatchAs case will invoke an error message.

# Details

Added one new function to function_codegen.py
visit_MatchAs, if there is a wildcard, create an 'otherwise' case in latex, else throw error

Added two new test cases to function_codegen_test.py
test_matchvalue_matchas_wildcard, tests if an 'otherwise' case is created in latex
test_matchvalue_matchas_nonempty, tests if an error is thrown

# References

#93

# Blocked by

#146


juliawgraham:
We reran all of the CI tests for this PR and it all seems to pass

odashi:
Okay will check it.

odashi:
It looks several refactoring can be still applied before merging. As this is the blocker of a broader refactoring (#155), I will process it on behalf of you.

odashi (approved):

Issue #83, opened 2022-11-10
Showing some special arrays


odashi:
Ref: #78 

We can also support showing some special array functions as a fixed, well-known form. For example, `ndarray.zeros((2, 3))` --> $\mathbf{0}$ or $\mathbf{0}^{2 \times 3}$

Casper-Guo:
Similar:

`np.eye(N, M=None, k=0)` --> $I_N$

`np.identity(N)` --> $I_N$

Casper-Guo:
I will do this once i figure out CI/CD on my other PR.

Once NDArray support is done, we can do the same for `np.transpose()` and `np.inv()` etc. as well

Casper-Guo:
What is the best place to check for function names for this issue? Should this be done in the `function_expander.py` file following #132 or should it be checked in `visit_Call` in `function_codegen.py` following #118 ?

odashi:
@Casper-Guo If you are planning to implement this on `FunctionExpander`, `function_expander_test.py` is the best place I think.

Casper-Guo:
Sorry I didn’t meant checking as in writing test cases. I was asking where best to extract numpy function names and generate the appropriate Latex. The two linked PRs seem to do this in different files

LakeBlair:
@odashi I just created a pull request for this feature. I wasn't too sure the best place to implement this feature so I added them to visit_call() in function_codegen.py. Could you review?

odashi:
@LakeBlair I checked it now. Overall it looks great, but it also requires some refactoring to support complete syntax and error handling. If you are okay, I will fix remaining things on behalf of you.

LakeBlair:
@odashi I added a generalization for np.zeros function, please check my most recent commit. Additionally, for your second bullet point
> I think these processes don't work if the subtree has unexpected syntax. It usually happens when users gave other functions with the same name. As the AST varies, we basically need complete check of the underlying structure of the given subtree.

I‘d love to implement a complete check for the subtree structure. Do you know in which file(s) I can find a similar example for such implementation?

LakeBlair:
@odashi I added a generalization for np.zeros function, please check my most recent commit. Additionally, for your second bullet point
> I think these processes don't work if the subtree has unexpected syntax. It usually happens when users gave other functions with the same name. As the AST varies, we basically need complete check of the underlying structure of the given subtree.

I‘d love to implement a complete check for the subtree structure. Do you know in which file(s) I can find a similar example for such implementation?

[...]

PR #148, opened 2022-12-04
Special array representation

LakeBlair:
# Overview
Added special array representation for np.zeros and np.identity.

# Details
image

# References
#83 

# Blocked by


examples/examples.ipynb
```diff
@@ -2,41 +2,20 @@
  "cells": [
```
odashi, line 2: (action not required) I guess this notebook file is no longer necessary as we could provide a comprehensive examples in Google Colab. I will remove this file later.

src/latexify/codegen/function_codegen.py
```diff
@@ -356,6 +356,17 @@ def visit_Call(self, node: ast.Call) -> str:
             (default_func_str + r"\mathopen{}\left(", r"\mathclose{}\right)"),
         )
 
+        if func_name == "zeros":
```
odashi, line 359: - For `zeros`, We don't need to constrain only a particular number of dimensions.
- I think these processes don't work if the subtree has unexpected syntax. It usually happens when users gave other functions with the same name. As the AST varies, we basically need complete check of the underlying structure of the given subtree.

odashi (commented):
Thanks! It looks this change needs some refactoring to support the feature and error handling completely. Which would you like either:

- fix them by yourself (in this case I will put comprehensive comments)
- delegate remaining tasks to me (your contribution is still recorded correctly)

odashi:
@LakeBlair Let's continue developing this pull request as you preferred in another thread.

Before providing additional comments, please resolve the following issues:

- Merge main into this branch and resolve all conflicts.
- run `./check.sh` and resolve all errors.

LakeBlair:
> @LakeBlair Let's continue developing this pull request as you preferred in another thread.
> 
> Before providing additional comments, please resolve the following issues:
> 
> * Merge main into this branch and resolve all conflicts.
> * run `./check.sh` and resolve all errors.

Just did both. Check if they look good to you.

src/latexify/codegen/function_codegen.py
```diff
@@ -425,7 +425,26 @@ def visit_Call(self, node: ast.Call) -> str:
...
+            str = ""
+            open_bracket = "{"
+            close_bracket = "}"
+            for i, elt in enumerate(node.args[0].elts):
+                str += "{" + self.visit(elt) + "}"
+                if i != len(node.args[0].elts) - 1:
+                    str += " " + r"\times" + " "
+
+            matrix_str = "0^" + open_bracket + str + close_bracket
+            return matrix_str
```
odashi, line 440: General suggestions:

- Requires complete syntax checking.
- Don't use `str` as a variable as it is reserved by the builtin type name. Overwriting builtins will confuse the behavior of the code (even if the current code doesn't involve any errors, it will happen in the future). Use `latex` instead if you need to store some generated strings.
- Suppress string concatenation as [it is significantly expensive](https://stackoverflow.com/a/3055541). Making a sequence (generator), then join all of them is generally better.
- I think `\mathbf{0}` ( $\mathbf{0}$ ) should be used to distinguish that this is not a scalar $0$.

```suggestion
            if len(node.args) != 1:
                # fall back to the default.
                
            arg0 = node.args[0]
                
            if not isinstance(arg0, ast.Tuple):
                # fall back to the default.
                # Tecunically we don't support `zeros(n)` where `n` is a scalar.
            
            dims_latex = r" \times ".join(self.visit(x) for x in arg0.elts)
            return fr"\mathbf{{0}}^{{{dims_latex}}}"
```

src/latexify/codegen/function_codegen_test.py
```diff
@@ -862,6 +862,34 @@ def test_use_set_symbols_compare(code: str, latex: str) -> None:
...
+    tree = ast.parse(
+        textwrap.dedent(
+            """
+        def f(a, b):
+            return np.zeros((a,b))
+            """
+        )
+    ).body[0]
+    latex = "f(a, b) = 0^{{a} \\times {b}}"
+    assert isinstance(tree, ast.FunctionDef)
```
odashi, line 875: We don't need complete parsing. It is enough to obtain only `ast.Call`.
```suggestion
    tree = ast_utils.parse_expr("zeros((a, b))")
    assert isinstance(tree, ast.Call)
```

src/latexify/codegen/function_codegen_test.py
```diff
@@ -862,6 +862,34 @@ def test_use_set_symbols_compare(code: str, latex: str) -> None:
     assert function_codegen.FunctionCodegen(use_set_symbols=True).visit(tree) == latex
 
 
+def test_generate_numpy_zeros():
```
odashi, line 865: Please write comprehensive tests to cover every edge cases, as in other parametric tests.

src/latexify/codegen/function_codegen_test.py
```diff
@@ -862,6 +862,34 @@ def test_use_set_symbols_compare(code: str, latex: str) -> None:
...
+            return np.zeros((a,b))
+            """
+        )
+    ).body[0]
+    latex = "f(a, b) = 0^{{a} \\times {b}}"
+    assert isinstance(tree, ast.FunctionDef)
+    assert FunctionCodegen().visit(tree) == latex
+
+
+def test_generate_numpy_identity():
```
odashi, line 879: ditto

src/latexify/codegen/function_codegen.py
LakeBlair: @odashi I made some changes to my code. 

- np.zeros and np.identity are now in a new function, treated as special numpy methods.
- I added some if statements checking for valid ast inputs. 
- I added some parametric tests. 

Some more considerations:

- Some string concatenation is simplified, some could be simplified further.
- In the case of np.zeros(0), I just represent it as $\mathbf{0}^{{1} \times {0}}$. There might be better ways of expressing it as a special case. Do you have any suggestions?

Let me know what you think.

src/latexify/codegen/function_codegen.py
```diff
@@ -416,6 +416,49 @@ def generate_matrix_from_array(data: list[list[str]]) -> str:
 
         return generate_matrix_from_array(rows)
 
+    def _generate_special_latex(self, node: ast.Call, func_name) -> str | None:
```
odashi, line 419:
```suggestion
    def _generate_special_latex(self, node: ast.Call, func_name: str) -> str | None:
```

odashi (commented):
Sorry I still think this change requires several refactoring, I will do it by my side.

odashi (approved):

PR #150, opened 2022-12-05
Fix bug and remove unnecessary brackets.

odashi:
# Overview

- Fix bug on the stop parameter of sum/prod (with refactoring of `_reduce_stop_parameter`).
- Removes unnecessary brackets from some expressions.

# Details

- The original implementation of `_reduce_stop_parameter` accidentally removes `-1` for some cases. This change replaces this function to return modified AST with more simpler rule, covering all cases correctly.

# References

- NA

# Blocked by

- NA


Casper-Guo:
Two quick questions about this PR:

1, Was the bug introduced by my PR (#115)? What is an example of a failing test / wrong behavior?

2, Looks like you made the function a lot less verbose by only composing the new AST node instead of the entire Latex string, is that correct?

odashi:
> Was the bug introduced by my PR (https://github.com/google/latexify_py/pull/115)? What is an example of a failing test / wrong behavior?

Yes. The erroneous case was this (and a similar statement in another block):

https://github.com/google/latexify_py/blob/eebbae799f9a36d88c0fb8c0c292db2097abcd37/src/latexify/codegen/function_codegen.py#L687

This statement returns `x + y` if the stop param is `x + y`, but it should be `x + y - 1`. This pull request also adds tests to capture this behavior.

> Looks like you made the function a lot less verbose by only composing the new AST node instead of the entire Latex string, is that correct?

It should be correct so long as I wrote it correctly.

odashi:
The new implementation does not implement any visitor that produces other than str. `_reduce_stop_parameter` is only the inner function of the class and can produce any values (but I also guess this kind of code should be provided as another library, meaning that we would need further refactoring after this PR)

ShigekiKarita (approved):
Nice to see many test cases.

odashi:
Yeah test is fast enough (it is 25 times faster than installing deps :)

PR #151, opened 2022-12-06
Revert BuiltinFnName

odashi:
# Overview

This change reverts `BuiltinFnName` enum.

CC: @ZibingZhang

# Details

`BuiltinFnName` was introduced by #130 to constrain keys in `BUILTIN_FUNCS`. I originally thought that it was useful, but after several development I realized that using raw string is more desirable:

- Enum introduces a strong type constraint. If I introduced mypy in my local environment, `BUILTIN_FUNCS` rejects all strings that are not defined in the `BuiltinFnName`, which is not a desirable behavior.
- We are required to write very long ientifiers, e.g., `constants.BuiltInFnName.EXP.value` rather than `"exp"`. The identifier also involves the same string `EXP` with the function name, meaning that we eventually need to write the function name with longer prefix every time.
- The enum is rarely used: in many cases the current implementation keeps the string representation.

# References

- NA

# Blocked by

- NA

### DIFF
--- a/src/latexify/constants.py
+++ b/src/latexify/constants.py
@@ -6,43 +6,6 @@ import dataclasses
 import enum
 
 
-class BuiltinFnName(str, enum.Enum):
-    """Built-in function name."""
-
-    ABS = "abs"
-    ACOS = "acos"
-    ACOSH = "acosh"
-    ARCCOS = "arccos"
-    ARCCOSH = "arcosh"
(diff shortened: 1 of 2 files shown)


ZibingZhang (approved):
LGTM

ShigekiKarita (approved):
Nice simplification.

Issue #57, opened 2022-10-21
Generates algorithmic environment


odashi:
I have worked recently to investigate if we could generate the LaTeX algorithm environment from the function.
Actually it can work like this:

<img width="353" alt="FfJj0MyVEAAe01M" src="https://user-images.githubusercontent.com/1023695/197110762-b71c1d0a-6bcd-4f95-b21f-15ec982d8ed4.png">

I think this feature is really useful for almost arbitrary functions.

odashi:
There are several caveats.

Specifically, IPython supports only MathJax, meaning that it accepts only a single expression. To output more complicated LaTeX document, we need to complie the source directly, and convert the generated file to an image that IPython accepts.

In the work above, I used pdflatex and pdf2image. They requires to install external libraries, e.g., texlive with the algorithmicx package and Poppler.

odashi:
An option to implement it for IPython is to use immediate newline in the single expression with appropriate manual spacing, but it requires some hacks. I think it would be better to apply external commands to generate the compiled image.

luisArandas:
Can this, if implemented, be added to the `ipynb` example?

odashi:
@luisArandas Yes, the poc works on Jupyter and/or Colab and we can add examples.

odashi:
I guess composing everything into `array` would be another option to show the result on Jupyter. Maybe we can introduce multiple codegen (one for algorithmic, one for jupyter).

<img width="651" alt="FgdCA52aMAA5ZQX" src="https://user-images.githubusercontent.com/1023695/200437862-3851774e-c991-4380-adbb-622a981858d4.png">

ZibingZhang:
I'd agree with multiple code gen, targeting either Jupyter. I'd be interested in taking a look into algorithmic at some point.

odashi:
I have some ideas about this.

- First, we can provide the `ExpressionCodegen`, which converts `ast.expr` to LaTeX. I think `ast.expr` is the minimal fragment of AST that we can share the same behavior among different output styles.
- Then, we provide several codegens, such as `FunctionCodegen`, `AlgorithmCodegen`, or something else, to process the whole syntax. They invokes `ExpressionCodegen` internally to obtain the underlying expressions, and organize all results along with the style that the codegen finally produces.

This design has also an advantage about programmatic use of this library. We can define a function `latexify.compile_expression("...expr...")`, which directly invokes `ExpressionCodegen` and returns the LaTeX. (also discussed in #108)

[...]

PR #152, opened 2022-12-06
Generates algorithmic environment from algpseudocode

ZibingZhang:
# Overview

Generates algorithmic environment from algpseudocode.

# Details

Creates a new class `algorithm_codegen` that should generate $\LaTeX$ code in the following form:

## Example

```python
    def collatz(n):
        iterations = 0
        while n > 1:
            if n % 2 == 0:
                n = n / 2
            else:
                n = 3 * n + 1
            iterations = iterations + 1
        return iterations
```

image

```LaTeX
\begin{algorithmic} ... \end{algorithmic}
```

Things to take into consideration:

* The configs won't be the same for this and `function_codegen`, at least for the time being. Namely `use_signature` will be disregarded if set.
* For the time being, this functionality is exposed through `get_latex(style=Style.ALGORITHM`, but there should be a way to use it as `@latexify.algorithm`. But what should the return type be? `str`? `LatexifiedFunction`?
* There are multiple algorithmic environments, how should be specify which ones to target? Which ones should be supported?

# References

Described by this issue: https://github.com/google/latexify_py/issues/57

# Blocked by

None

### DIFF
--- a/src/integration_tests/algorithmic_style_test.py
+++ b/src/integration_tests/algorithmic_style_test.py
@@ -1,3 +1,7 @@
+"""End-to-end test cases of algorithmic style."""
+
+from __future__ import annotations
+
 from typing import Any, Callable
 
 from latexify import frontend
--- a/src/integration_tests/algorithmic_style_test.py
+++ b/src/integration_tests/algorithmic_style_test.py
@@ -35,11 +35,13 @@ def test_factorial() -> None:
 
     latex = (
         r"\begin{algorithmic} "
+        r"\Procedure{fact}{$n$} "
         r"\If{$n = 0$} "
         r"\State \Return $1$ "
         r"\Else "
         r"\State \Return $n \mathrm{fact} \mathopen{}\left( n - 1 \mathclose{}\right)$ "
         r"\EndIf "
+        r"\EndProcedure "
         r"\end{algorithmic}"
     )
     check_algorithm(fact, latex)
@@ -58,6 +60,7 @@ def test_collatz() -> None:
 
     latex = (
         r"\begin{algorithmic} "
+        r"\Procedure{collatz}{$n$} "
         r"\State $\mathrm{iterations} \gets 0$ "
         r"\While{$n > 1$} "
         r"\If{$n \mathbin{\%} 2 = 0$} "
@@ -67,6 +70,7 @@ def test_collatz() -> None:
         r"\State $\mathrm{iterations} \gets \mathrm{iterations} + 1$ "
         r"\EndWhile "
         r"\State \Return $\mathrm{iterations}$ "
+        r"\EndProcedure "
         r"\end{algorithmic}"
     )
 
--- a/src/integration_tests/algorithmic_style_test.py
+++ b/src/integration_tests/algorithmic_style_test.py
@@ -75,5 +75,4 @@ def test_collatz() -> None:
         r"\EndProcedure "
         r"\end{algorithmic}"
     )
-
     check_algorithm(collatz, latex)
--- a/src/integration_tests/algorithmic_style_test.py
+++ b/src/integration_tests/algorithmic_style_test.py
@@ -26,6 +26,8 @@ def check_algorithm(
     assert latexified == latex
 
 
+# TODO(ZibingZhang) changing fact -> factorial breaks because factorial is replaced by
+#  !, substitutions should not occur for the name of the procedure
 def test_factorial() -> None:
     def fact(n):
         if n == 0:
--- a/src/integration_tests/algorithmic_style_test.py
+++ b/src/integration_tests/algorithmic_style_test.py
@@ -27,8 +27,6 @@ def check_algorithm(
     assert latexified == latex
 
 
-# TODO(ZibingZhang) changing fact -> factorial breaks because factorial is replaced by
-#  !, substitutions should not occur for the name of the procedure
 def test_factorial() -> None:
     def fact(n):
         if n == 0:
--- a/src/integration_tests/algorithmic_style_test.py
+++ b/src/integration_tests/algorithmic_style_test.py
@@ -1,0 +1,69 @@
+from typing import Any, Callable
+
+from latexify import frontend
+
+
+def check_algorithm(
+    fn: Callable[..., Any],
+    latex: str,
+    **kwargs,
+) -> None:
+    """Helper to check if the obtained function has the expected LaTeX form.
+    Args:
+        fn: Function to check.
+        latex: LaTeX form of `fn`.
+        **kwargs: Arguments passed to `frontend.get_latex`.
+    """
+    # Checks the syntax:
+    #     def fn(...):
+    #         ...
+    #     latexified = get_latex(fn, style=ALGORITHM, **kwargs)
+    latexified = frontend.get_latex(fn, style=frontend.Style.ALGORITHMIC, **kwargs)
+    assert latexified == latex
+
+
+def test_factorial() -> None:
+    def fact(n):
+        if n == 0:
+            return 1
+        else:
+            return n * fact(n - 1)
+
+    latex = (
+        r"\begin{algorithmic} "
+        r"\If{$n = 0$} "
+        r"\State \Return $1$ "
+        r"\Else "
+        r"\State \Return $n \mathrm{fact} \mathopen{}\left( n - 1 \mathclose{}\right)$ "
+        r"\EndIf "
+        r"\end{algorithmic}"
+    )
+    check_algorithm(fact, latex)
+
+
+def test_collatz() -> None:
+    def collatz(n):
+        iterations = 0
+        while n > 1:
+            if n % 2 == 0:
+                n = n / 2
+            else:
+                n = 3 * n + 1
+            iterations = iterations + 1
+        return iterations
+
+    latex = (
+        r"\begin{algorithmic} "
+        r"\State $\mathrm{iterations} \gets 0$ "
+        r"\While{$n > 1$} "
+        r"\If{$n \mathbin{\%} 2 = 0$} "
+        r"\State $n \gets \frac{n}{2}$ "
+        r"\Else \State $n \gets 3 n + 1$ "
+        r"\EndIf "
+        r"\State $\mathrm{iterations} \gets \mathrm{iterations} + 1$ "
+        r"\EndWhile "
+        r"\State \Return $\mathrm{iterations}$ "
+        r"\end{algorithmic}"
+    )
+
+    check_algorithm(collatz, latex)
--- a/src/integration_tests/algorithmic_style_test.py
+++ b/src/integration_tests/algorithmic_style_test.py
@@ -13,6 +13,7 @@ def check_algorithm(
     **kwargs,
 ) -> None:
     """Helper to check if the obtained function has the expected LaTeX form.
+
     Args:
         fn: Function to check.
         latex: LaTeX form of `fn`.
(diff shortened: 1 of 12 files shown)


ZibingZhang:
@odashi let me know what you think about this approach. Looking for a bit of early feedback before continuing.

odashi:
@ZibingZhang Could you open pull requests instead of leaving it as draft? I'm using PR's status to manage my taskboard and sometimes draft PRs are overlooked.

odashi:
Thanks! I will take a look at the code later today. Some comments about the description:

> `else n % 2 == 1:`

This may be a typo.

> `@latexify.algorithm`

If we provide this function, it must provide a way to correctly print the expression onto Jupyter. Since algorithm environmets are not supported by Jupyter (MathJax), we need to simulate it by ourselves.

This means that the algorithm version of "LatexifiedFunction" needs two codegens for `__str__` and `_repr_latex_`, but this is far beyond the scope of this PR IMO.

> Which environment

`algorithmic` is fine I think. It is okay to start considering to support other envs when they are actually requested.

At this point, I guessed the enum value should be `ALGORITHMIC` rather than `ALGORITHM` to avoid confusion in the future.

src/latexify/codegen/algorithm_codegen.py
```diff
@@ -0,0 +1,67 @@
+import ast
+
+from latexify import exceptions
+from latexify.codegen import function_codegen
+
+
+class AlgorithmCodegen(function_codegen.FunctionCodegen):
```
odashi, line 7: Just a quick comment: This class shouldn't inherit `FunctionCodegen` to avoid a weird "is-a" class relationship, e.g., `isinstance(AlgorithmCodegen(), FunctionCodegen)` should be `False`. Inheritance also brings all functionality in `FunctionCodegen` implicitly, which may increase the maintenance cost in the future.

`FunctionCodegen` can be instantiated as a private member of this class (aka composition) and used internally only when we need to delegate the codegen logic to it.

ZibingZhang:
> used internally only when we need to delegate the codegen logic to it

I'm not sure if this is possible, due to the recursive nature of ASTs. If you look at the first commit I had something like

```python
def visit_SomeNode(node):
    self._function_codegen.visit(node)
```

But something of this nature won't work since recursive calls will end up in the FunctionCodegen methods instead of back in the AlgorithmCodegen methods.

odashi:
> I'm not sure if this is possible

It should work as far as the subtree processing is completely delegated. Since both `NodeVisitor` and our `FunctionCodegen` is stateless and does not change the given AST, we can invoke `visit` as a usual function.

Unfortunately there is no private class in Python, and many techniques available in other languages (such as C++) can't be used.

ZibingZhang:
> It should work as far as the subtree processing is completely delegated. Since both `NodeVisitor` and our `FunctionCodegen` are stateless and don't change the given AST, we can invoke `visit` as a usual function.

Consider `visit_Attribute`:

```python
class ACodegen:
    def visit_Attribute(self, node: ast.Attribute) -> str:
        vstr = self.visit(node.value)
        astr = self._identifier_converter.convert(node.attr)[0]
        return vstr + "." + astr
    
    def visit_If(self, node: ast.If) -> str:
        raise Error("Unsupported")

class BCodegen:
    def visit_Attribute(self, node: ast.Attribute) -> str:
        return self._a_codegen.visit(node)

    def visit_If(self, node: ast.If) -> str:
        return "..."
```

If we have some structure like:

```json
{ "attribute": { "value": { "if": "..." } }
```

The call stack would look like

```text
BCodegen.visit_Attribute 
-> ACodegen.visit_Attribute
-> ACodegen.visit_If
-> Error
```

whereas the desired behavior would be

```text
BCodegen.visit_Attribute 
-> ACodegen.visit_Attribute
-> BCodegen.visit_If
```

odashi:
In the case above doesn't follow the rule "the subtree processing is completely delegated.".

Before continuing discussion, we need to understand how the Python AST is constructed. Overall, all ASTs look like as follows:

`[mod subtree] -> [stmt subtree] -> [expr subtree]`

where `stmt` subtree never generate `mod` and `expr` never generate `stmt`. ([the official guidance](https://docs.python.org/3/library/ast.html#abstract-grammar))

In `FunctionCodegen` and `AlgorithmCodegen`, we need to handle `stmt` and `expr` to generate the final LaTeX, but once the codegen walks into `expr`, the subroutine processes only `expr` and it never meet any `stmt` branch in this recursion. That means we can completely separate `expr` processing into another library, and this is why I proposed `ExpressionCodegen` in #57.

In this PR, `AlgorithmCodegen` need to implement every visitors for `stmt` nodes by itself, but can delegate every processing for `expr` to another. Since the current codebase implements every rule in `FunctionCodegen`, we can simply delegate everything to it, and we don't need additional information to invoke `expr` visitors in `FunctionCodegen`.

odashi:
The contrast between `stmt` and `expr` is important I guess, because codegen rules to control the style (the appearance of the overall algorithm) is basically implemented by only `stmt` rules.

ZibingZhang:
That makes sense ty for explaining!

The only issue I can maybe see down the line if `AlgorithmCodegen` decides that an expression should take one form but `FunctionCodegen` decides it should take another, but I can't think of any examples off the top of my head, so maybe it's a non-issue.

odashi:
> but it's an ast.stmt

I think it's `AST`, not `stmt`. This is because the class is used to aggregate a set of rules (in this case, a sub-expression of a comprehension in some other expr), and appears only within a limited context. This kind of subtree can't be treated as an `expr`, e.g. `a + <comprehension>` is illegal.

We should be able to treat this kind of subtrees safely as a part of either `stmt` or `expr` because they shouldn't violate the entire rule of the AST: `stmt -> expr`. In this case it is okay to assume that this is `expr`.

odashi:
ExpressionCodegen should be separated into other pull request since this is an independent feature.

ZibingZhang:
> ExpressionCodegen should be separated into other pull request since this is an independent feature.

Sounds good. Will tackle this within the next few days!

odashi (commented):
Dummy, please re-request afterwards

src/integration_tests/integration_utils.py
```diff
@@ -1,3 +1,7 @@
+"""Utilities for integration tests."""
```
ZibingZhang, line 1: Not great to have file named `utils.py`, since we have so many `...utils.py` files.

src/latexify/codegen/algorithmic_codegen_test.py
```diff
@@ -0,0 +1,125 @@
...
+    assert isinstance(node, ast.While)
+    assert algorithmic_codegen.AlgorithmicCodegen().visit(node) == latex
+
+
+def test_visit_while_with_else() -> None:
+    with pytest.raises(exceptions.LatexifyNotSupportedError):
+        node = ast.parse(
+            textwrap.dedent(
+                """
+                while True:
```
ZibingZhang, line 117: Fixed whitespacing issues in `function_codegen_test.py` to match this style, don't know what I was doing when I wrote those tests...

src/latexify/frontend.py
```diff
@@ -173,7 +190,7 @@ def expression(
     This function is a shortcut for `latexify.function` with the default parameter
     `use_signature=False`.
     """
-    kwargs["use_signature"] = kwargs.get("use_signature", False)
```
ZibingZhang, line 176: Do we still need `use_signature` if we have `expression`?

src/latexify/codegen/algorithmic_codegen.py
```diff
@@ -0,0 +1,109 @@
...
+    def visit_Assign(self, node: ast.Assign) -> str:
+        """Visit an Assign node."""
+        operands: list[str] = [
+            self._expression_codegen.visit(target) for target in node.targets
+        ]
+        operands.append(self._expression_codegen.visit(node.value))
+        operands_latex = r" \gets ".join(operands)
+        return rf"\State ${operands_latex}$"
+
+    def visit_Expr(self, node: ast.Expr) -> str:
```
ZibingZhang, line 51: Required for something like `if True: x`. Have also added `visit_Expr` to `expression_codegen.py`.

src/integration_tests/algorithmic_style_test.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    # Checks the syntax:
+    #     def fn(...):
+    #         ...
+    #     latexified = get_latex(fn, style=ALGORITHM, **kwargs)
+    latexified = frontend.get_latex(fn, style=frontend.Style.ALGORITHMIC, **kwargs)
+    assert latexified == latex
+
+
+# TODO(ZibingZhang) changing fact -> factorial breaks because factorial is replaced by
```
ZibingZhang, line 29: Need to be careful about how we're substituting functions

src/latexify/codegen/algorithmic_codegen.py
odashi: `Expr` is `stmt` that owns only one expression. I think it should be procecced by `stmt` codegen.

ZibingZhang: ah yes you're right, fixing that

src/latexify/frontend.py
```diff
@@ -173,7 +190,7 @@ def expression(
...
+    kwargs["style"] = Style.EXPRESSION
```
odashi, line 193: This function no longer need to invoke `function`, just invoking `get_latex` directly is fine.

src/latexify/codegen/algorithmic_codegen.py
```diff
@@ -0,0 +1,109 @@
...
+    def visit_Module(self, node: ast.Module) -> str:
+        """Visit a Module node."""
+        return self.visit(node.body[0])
+
+    def visit_Return(self, node: ast.Return) -> str:
+        """Visit a Return node."""
+        return (
+            rf"\State \Return ${self._expression_codegen.visit(node.value)}$"
+            if node.value is not None
+            else rf"\State \Return ${codegen_utils.convert_constant(None)}$"
```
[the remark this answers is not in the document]
ZibingZhang: agreed

src/integration_tests/algorithmic_style_test.py
```diff
@@ -0,0 +1,79 @@
+"""End-to-end test cases of algorithmic style."""
+
+from __future__ import annotations
+
+from typing import Any, Callable
+
+from latexify import frontend
```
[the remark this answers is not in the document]
ZibingZhang: I can move it, but `function_expansion_test` and `regression_test` basically use `frontend` directly, as they use `frontend.function`.

Similarly, this file uses `function.get_latex`. Both of these functions are exported in `__init__.py`, so it feels somewhat like an integration test, because we're testing end-to-end behavior from function to generated $LaTeX$ output

src/latexify/codegen/algorithmic_codegen.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    def visit_While(self, node: ast.While) -> str:
+        """Visit a While node."""
+        if node.orelse:
+            raise exceptions.LatexifyNotSupportedError(
+                "Codegen does not support while statements with an else clause."
```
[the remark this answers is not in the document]
ZibingZhang: That's fair, was trying to model it off the error message in `function_codegen` about only having assign nodes in multiline functions. Unsure how to change that message as well to be more user friendly

src/integration_tests/algorithmic_style_test.py
ZibingZhang: https://github.com/google/latexify_py/issues/160

src/latexify/frontend.py
ZibingZhang: `expression` returns a `LatexifiedFunction`, while `get_latex` returns `str`. I think we still need the extra work that `function` provides to wrap the result of `get_latex`, no?

odashi: Ah sorry I was a bit confused! I will consider these implementations later.

src/integration_tests/algorithmic_style_test.py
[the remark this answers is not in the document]
odashi: So it turns that the issue comes from the old implementation of these tests, will consider it later.

odashi (approved):
Maybe good to merge, thanks!

PR #153, opened 2022-12-06
Enable mypy

odashi:
# Overview

Introduces mypy as a new CI. Also fixes several typing errors discovered by the non-strict mode of mypy.

This change removes the deprecated `with_latex`, which is no longer useful to support (especially with  correct type hints).

# References

NA

# Blocked by

- #151

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,3 +69,17 @@ jobs:
           python -m pip install isort
       - name: Check
         run: python -m isort --check src
+  mypy:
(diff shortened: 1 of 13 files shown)


odashi:
CI is still failing, will fix it

ShigekiKarita (approved):
Very nice. For my education, is there any reason why you choose mypy rather than pytype, pyright?

odashi:
It is because mypy is provided by the official, and has more audience than others. Other tools are also good though; so this is just kinda bike-shedding decision.

PR #158, opened 2022-12-09
MatchOr and guard statements

erica-w-fu:
MatchOr and MatchAnd and implemented and tested.

# Overview

All of these produce the correct latex code when a user enters a MatchOr case or guard statements in a python match statement. 

# Details

made edits to visit_Match
- added guard statement case, calls visit_Compare
- replaced any instances of "subject_name" with the subject name (ie if someone write match x, x will be the subject name)

created a new function visit_MatchOr
- will add "\lor" between each of the patterns
- will invoke self.visit on each pattern

We worked with four collaborators
@erica-w-fu
@Lucybean-hi
@juliawgraham
@Yuqi07

# References

#93

# Blocked by

#146
#147


odashi:
Thanks, could you follow the recent changes in main?

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,63 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
...
+            )
+
+    def visit_MatchOr(self, node: ast.MatchOr) -> str:
+        """Visit a MatchOr node"""
+        latex = ""
+        for i, pattern in enumerate(node.patterns):
+            if i != 0:
+                latex += r" \lor " + self.visit(pattern)
+            else:
+                latex += self.visit(pattern)
```
odashi, line 593: Don't use `+=` for string concatenation in for-loops. generate a list first, then join them with intermediate operators.

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,63 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
...
+                    "Multiple statements are not supported in Match nodes."
+                )
+            true_latex = self.visit(match_case.body[0])
+            cond_latex = self.visit(match_case.pattern)
+
+            if i < len(node.cases) - 1:  # no wildcard cases
+                if match_case.guard:
+                    cond_latex = self.visit(match_case.guard)
+                    subject_latex = ""  # getting variable from cond_latex
+                if not cond_latex:
```
odashi, line 555: Don't use generated LaTeX for any conditions. It is not stable and the value may change in the future. You need to always check the original ast for conditioning.

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,63 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
...
+        for i, match_case in enumerate(node.cases):
+            if len(match_case.body) != 1:
+                raise exceptions.LatexifySyntaxError(
+                    "Multiple statements are not supported in Match nodes."
+                )
+            true_latex = self.visit(match_case.body[0])
+            cond_latex = self.visit(match_case.pattern)
+
+            if i < len(node.cases) - 1:  # no wildcard cases
+                if match_case.guard:
```
odashi, line 552: Please use always `x is None` syntax if none checking is required.

odashi (commented):
Only general comments

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,63 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
...
+
+    def visit_MatchValue(self, node: ast.MatchValue) -> str:
+        """Visit a MatchValue node"""
+        latex = self.visit(node.value)
+
+        return "subject_name = " + latex
+
+    def visit_MatchAs(self, node: ast.MatchAs) -> str:
+        """Visit a MatchAs node"""
+        """If MatchAs is a wildcard, return 'otherwise' case, else throw error"""
```
ZibingZhang, line 578: nit: should be combined with the line above to form a single, multiline docstring

src/latexify/codegen/function_codegen.py
```diff
@@ -536,6 +536,63 @@ def visit_IfExp(self, node: ast.IfExp) -> str:
         latex += self.visit(node)
         return latex + r", & \mathrm{otherwise} \end{array} \right."
 
+    def visit_Match(self, node: ast.Match) -> str:
+        """Visit a match node"""
```
ZibingZhang, line 540: nit: summary line should be a complete sentence (missing a period)

src/latexify/codegen/function_codegen_test.py
```diff
@@ -744,3 +744,284 @@ def test_use_set_symbols_compare(code: str, latex: str) -> None:
     tree = ast.parse(code).body[0].value
     assert isinstance(tree, ast.Compare)
     assert function_codegen.FunctionCodegen(use_set_symbols=True).visit(tree) == latex
+
+
+@test_utils.require_at_least(10)
+def test_matchvalue() -> None:
```
ZibingZhang, line 750: Shouldn't these tests belong in `function_codegen_match_test.py`?

nit: most test methods in this file have the form `test_visit_XYZ_explanation`

erica-w-fu:
> Thanks, could you follow the recent changes in main?

Could you clarify what those recent changes are?

ZibingZhang:
> Could you clarify what those recent changes are?

Your branch is behind the main branch. Can be resolved by merging main into this branch and resolving conflicts.

src/latexify/codegen/function_codegen.py
Yuqi07: Could you please clarify what does "check the original ast for conditioning"? Does it mean that we should parse to get the subject_latex ourselves, rather than getting that from cond_latex?

odashi: `node` passed to each method has sufficient information to process all codegen rules. The code has to check the inner structure of `node` rather than checking the generated latex.

src/latexify/codegen/function_codegen.py
```diff
@@ -141,42 +141,99 @@ def visit_If(self, node: ast.If) -> str:
...
-                body_latex = self.visit(case.body[0])
-                cond_latex = self.visit(case.pattern)
-                case_latexes.append(
-                    body_latex + r", & \mathrm{if} \ " + subject_latex + cond_latex
+        latex = r"\left\{ \begin{array}{ll} "
+        subject_latex = self.visit(node.subject)
+        for i, match_case in enumerate(node.cases):
+            if len(match_case.body) != 1:
+                raise exceptions.LatexifySyntaxError(
+                    "Multiple statements are not supported in Match nodes."
```
ZibingZhang, line 149:
```suggestion
                    "Multiple statements are not supported in a Match case."
```

Related to [this comment](https://github.com/google/latexify_py/pull/152#discussion_r1045041035) I got, but talking about match case is probably more intuitive.

src/latexify/codegen/function_codegen.py
```diff
@@ -141,42 +141,99 @@ def visit_If(self, node: ast.If) -> str:
...
+        """Visit a MatchOr node"""
+        latex = ""
+        for i, pattern in enumerate(node.patterns):
+            if i != 0:
+                latex += r" \lor " + self.visit(pattern)
+            else:
+                latex += self.visit(pattern)
+        return latex
+
+    def _reduce_stop_parameter(self, node: ast.BinOp) -> str:
```
ZibingZhang, line 199: Accidentally left over from merge?

Lucybean-hi:
We updated functions based on your suggestions, resolved all conflicts with the current main, and fixed all CI errors. Please take a check.

src/latexify/codegen/function_codegen.py
```diff
@@ -162,21 +162,49 @@ def visit_Match(self, node: ast.Match) -> str:
...
                 )
 
-        return (
+        latex = (
             r"\left\{ \begin{array}{ll} "
             + r" \\ ".join(case_latexes)
             + r" \end{array} \right."
         )
 
+        latex_final = latex.replace("subject_name", subject_latex)
```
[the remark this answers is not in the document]
erica-w-fu: Yes that is something we considered as well and would love to hear suggestions.

Since each of the visit_Match... functions cannot access the subject_latex variable(each of those functions can only have the ast node and pattern as the input), we had to find a workaround.  We tried to use python format and % to insert the subject_latex but the use of % and {} in latex made this very difficult. Do you have any suggestions for how to better implement this (or possible a better name than "subject_name" to replace)

src/latexify/codegen/function_codegen.py
```diff
@@ -162,21 +162,49 @@ def visit_Match(self, node: ast.Match) -> str:
             if i < len(node.cases) - 1:
                 body_latex = self.visit(case.body[0])
                 cond_latex = self.visit(case.pattern)
-                case_latexes.append(
-                    body_latex + r", & \mathrm{if} \ " + subject_latex + cond_latex
-                )
+                if case.guard is not None:
+                    cond_latex = self._expression_codegen.visit(case.guard)
+                    subject_latex = ""  # getting variable from cond_latex
```
ZibingZhang, line 167: Does this line actually do anything? 

```
subject_latex = "asdf"  # getting variable from cond_latex
```

seems to not break any tests.

src/latexify/codegen/function_codegen.py
```diff
@@ -162,21 +162,49 @@ def visit_Match(self, node: ast.Match) -> str:
...
+
+                case_latexes.append(body_latex + r", & \mathrm{if} \ " + cond_latex)
             else:
                 case_latexes.append(
                     self.visit(node.cases[-1].body[0]) + r", & \mathrm{otherwise}"
```
ZibingZhang, line 172:
```suggestion
                    self.visit(case.body[0]) + r", & \mathrm{otherwise}"
```

src/latexify/codegen/function_codegen.py
ZibingZhang, line 167:
```suggestion
```

Can be safely deleted :)

src/latexify/codegen/function_codegen.py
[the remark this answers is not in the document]
ZibingZhang: Ah yeah I see the dilemma, bit perplexing lol. Not entirely sure how to solve but I'll give it some thought. Thanks for tackling `match` statements btw! Really cool feature :)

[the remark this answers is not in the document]
ZibingZhang: Not sure what @odashi thinks of this solution, but you could create a stack of `_subject_latexes: listr[str] = []`, and at the very beginning of `visit_Match` do something like

```python
subject_latex = self._expression_codegen.visit(node.subject)
self._subject_latexes.append(subject_latex)
```

and then right before you return do something like

```python
latex = (
    r"\left\{ \begin{array}{ll} "
    + r" \\ ".join(case_latexes)
    + r" \end{array} \right."
)
        
self._subject_latexes.pop()
return latex
```

and you would also have

```
    def visit_MatchValue(self, node: ast.MatchValue) -> str:
        """Visit a MatchValue node."""
        latex = self._expression_codegen.visit(node.value)
        return self._subject_latexes[-1] + " = " + latex
```

A stack is only required if nested match statements is supported, which it seems like it is, although I have no idea if it's valid $\LaTeX$ output? In any case this situation I think necessitates a test for nested match statements (assuming they're supported), in order to test that in `visit_MatchValue`, the LHS takes the appropriate `_subject_latex`, i.e. the inner one should take the inner `subject` while the outer one takes the outer.

src/latexify/codegen/function_codegen_match_test.py
```diff
@@ -184,3 +184,112 @@ def test_matchvalue_mutliple_statements() -> None:
...
+    )
+
+
+@test_utils.require_at_least(10)
+def test_visit_match_case_with_if_and() -> None:
+    tree = ast.parse(
+        textwrap.dedent(
+            """
+            match x:
+                case x if x>0 and x<=10:
```
ZibingZhang, line 217:
```suggestion
                case x if x > 0 and x <= 10:
```

src/latexify/codegen/function_codegen_match_test.py
```diff
@@ -184,3 +184,112 @@ def test_matchvalue_mutliple_statements() -> None:
...
+    )
+
+
+@test_utils.require_at_least(10)
+def test_visit_matchcase_with_if_or() -> None:
+    tree = ast.parse(
+        textwrap.dedent(
+            """
+            match x:
+                case x if x>0 or x<=10:
```
ZibingZhang, line 239:
```suggestion
                case x if x > 0 or x <= 10:
```

src/latexify/codegen/function_codegen.py
```diff
@@ -162,21 +162,47 @@ def visit_Match(self, node: ast.Match) -> str:
             if i < len(node.cases) - 1:
```
ZibingZhang, line 162: According to the [grammar specs](https://docs.python.org/3/reference/grammar.html) we always have at least one `case`, so maybe we can split this for loop into two parts?

```python
for case in node.cases[:-1]:
    ...

case = node.cases[-1]
...
```

This way we can get rid of `enumerate`, as well as the `if i < len(node.cases) - 1` check in each iteration of the loop, which imo would make it a bit cleaner since we do something different for everything but the last element.

We would then also have to move the case validity check into a separate, "private" method.

odashi: It would be okay to apply it after this PR.

src/latexify/codegen/function_codegen.py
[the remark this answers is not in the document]
odashi: I think the @ZibingZhang 's solution is desirable at this point.

One minor point is that the private member `_subject_latexes` sounds weird for every other visitors, more specific name (like `_match_subject_stack`) would be better.

src/latexify/codegen/function_codegen.py
```diff
@@ -162,21 +162,47 @@ def visit_Match(self, node: ast.Match) -> str:
...
                 body_latex = self.visit(case.body[0])
                 cond_latex = self.visit(case.pattern)
-                case_latexes.append(
-                    body_latex + r", & \mathrm{if} \ " + subject_latex + cond_latex
-                )
+                if case.guard is not None:
+                    cond_latex = self._expression_codegen.visit(case.guard)
```
odashi, line 166: This doesn't generate correct expression: e.g., if we get `x, y if x == y`, it generates $\mathrm{if} \ x = y$, but $x$ and $y$ are not defined. It is better to avoid supports for guards from this PR and make additional discussions.

src/latexify/codegen/function_codegen.py
```diff
@@ -162,21 +162,47 @@ def visit_Match(self, node: ast.Match) -> str:
...
+
+    def visit_MatchOr(self, node: ast.MatchOr) -> str:
+        """Visit a MatchOr node."""
+        case_latexes = []
+        for i, pattern in enumerate(node.patterns):
+            if i == 0:
+                case_latexes.append(self.visit(pattern))
+            else:
+                case_latexes.append(r" \lor " + self.visit(pattern))
+        return "".join(case_latexes)
```
odashi, line 208: Too redundant. Joining all conditions with `\lor` is much efficient.
```suggestion
        return r" \lor ".join(self.visit(p) for p in node.patterns)
```

src/latexify/codegen/function_codegen_match_test.py
```diff
@@ -28,15 +28,15 @@ def f(x):
...
+        r"1, & \mathrm{if} \ x = 0 \\"
         r" 3 x, & \mathrm{otherwise}"
         r" \end{array} \right."
     )
     assert function_codegen.FunctionCodegen().visit(tree) == expected
 
 
 @test_utils.require_at_least(10)
-def test_matchvalue() -> None:
+def test_visit_match() -> None:
```
odashi, line 39: This name doesn't make sense because it looks too general.

src/latexify/codegen/function_codegen.py
```diff
@@ -162,21 +162,47 @@ def visit_Match(self, node: ast.Match) -> str:
...
+        Visit a MatchAs node.
+        If MatchAs is a wildcard, return 'otherwise' case, otherwise throw an error.
+        """
+        if node.pattern is None:
+            return ""
+        else:
+            raise exceptions.LatexifySyntaxError(
+                "Nonempty as-patterns are not supported in MatchAs nodes."
+            )
+
```
odashi, line 199: Remove this function. We don't need to process `MatchAs` explicitly, because:

- Wildcards are processed in `visit_Match`
- Any other cases raises an error.

```suggestion
```

src/latexify/codegen/function_codegen_match_test.py
```diff
@@ -184,3 +184,112 @@ def test_matchvalue_mutliple_statements() -> None:
...
+                    return 2
+            """
+        )
+    ).body[0]
+
+    assert (
+        function_codegen.FunctionCodegen().visit(tree)
+        == r"\left\{ \begin{array}{ll}"
+        + r"1, & \mathrm{if} \ x > 0 \\ "
+        + r"2, & \mathrm{otherwise} \end{array} \right."
```
odashi, line 207: As commented, this output is weird, and we shouldn't support guards at this point.

ZibingZhang:
@odashi if you're busy would you mind if i looked into this for merging? I think a lot of the conflicts were due to my changes

odashi:
@ZibingZhang Yeah please!

PR #161, opened 2022-12-10
Change default multiplication behavior

ZibingZhang:
# Overview

By default, multiplication should have a `\cdot`.

# Details

Based on [this discussion](https://github.com/google/latexify_py/issues/89#issuecomment-1344967636), it's clear that not having `\cdot` is the exception rather than the rule.

As a precursor to work implementing the exception behavior, we should first move the default behavior to include a `\cdot`, in order to make the future PR easier to follow. We basically split the that issue into two

1. first change default behavior
2. TODO: implement exceptions

# References

https://github.com/google/latexify_py/issues/89

# Blocked by

None


src/latexify/codegen/expression_codegen.py
```diff
@@ -122,7 +122,7 @@ class BinOpRule:
         operand_left=BinOperandRule(force=True),
         operand_right=BinOperandRule(wrap=False),
     ),
-    ast.Mult: BinOpRule("", " ", ""),
+    ast.Mult: BinOpRule("", r" \cdot ", ""),
```
odashi, line 125: Revise `MatMult` as well.

ZibingZhang: revised

odashi:
As per comments in #162, this pull request is merged after that change.

odashi (approved):

ZibingZhang:
Think this is good to merge as well

odashi (approved):

PR #162, opened 2022-12-10
Move definitions of expression rules.

odashi:
# Overview

This change doesn't introduce any new features.

This change moves all rule definitions for `ExpressionCodegen` to `expression_rules.py`.

CC @ZibingZhang 

# Details

# References

- https://github.com/google/latexify_py/pull/155#discussion_r1043383632

# Blocked by

NA


odashi:
ah it breaks blame. will fix it.

ZibingZhang:
I see you've marked as blocked by #161, but feel free to merge if working. I can resolve any conflicts in my branch later on.

src/latexify/codegen/expression_rules_test.py
```diff
@@ -0,0 +1,25 @@
...
+        (ast.Call(), expression_rules._CALL_PRECEDENCE),
+        (ast.BinOp(op=ast.Add()), expression_rules._PRECEDENCES[ast.Add]),
+        (ast.UnaryOp(op=ast.UAdd()), expression_rules._PRECEDENCES[ast.UAdd]),
+        (ast.BoolOp(op=ast.And()), expression_rules._PRECEDENCES[ast.And]),
+        (ast.Compare(ops=[ast.Eq()]), expression_rules._PRECEDENCES[ast.Eq]),
+        (ast.Name(), expression_rules._INF_PRECEDENCE),
+        (ast.Attribute(), expression_rules._INF_PRECEDENCE),
+    ],
+)
+def test_get_precedence(tree: ast.AST, precedence: int) -> None:
```
ZibingZhang, line 24: Half the tests call this argument `tree` and the other half `node`. Personally I prefer `node` but don't really care but would make it easier to understand tests if the same variable name was used across them.

ZibingZhang (approved):

src/latexify/codegen/expression_rules.py
```diff
@@ -0,0 +1,266 @@
...
+    """
+
+    left: str
+    right: str = ""
+    is_unary: bool = False
+    is_wrapped: bool = False
+
+
+# name => left_syntax, right_syntax, is_wrapped
+BUILTIN_FUNCS: dict[str, FunctionRule] = {
```
ZibingZhang, line 217: Assuming this constant doesn't have a `_` prefix since it's imported from another file. In that case, shouldn't `_MATH_SYMBOLS` in `math_symbols.py` also have it's `_` prefix removed?

odashi: moved `_MATH_SYMBOLS` to `expression_rules.MATH_SYMBOLS`

src/latexify/codegen/expression_rules_test.py
odashi: renamed to `node` anyway since the name have been used in codegens. `tree` is used in frontend though. I prefer this name a bit because it slightly means "entire tree" rather than the fragment. (But this is just a bike-shedding problem I guess)

ShigekiKarita (approved):

src/latexify/codegen/expression_rules_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+import ast
+
+import pytest
+
+from latexify.codegen import expression_rules
+
+
+@pytest.mark.parametrize(
+    "tree,precedence",
```
ShigekiKarita, line 13: tree -> node (or node -> tree)?

odashi: Oh yeah thanks!

PR #163, opened 2022-12-11
Algorithmic decorator (implement output for _repr_latex_)

ZibingZhang:
# Overview

Adds a `IPythonAlgorithmicCodegen` class which converts algorithms into $\LaTeX$ for IPython.

# References

continues work on https://github.com/google/latexify_py/issues/57

# Blocked by

None


ZibingZhang:
A point of interest is that the result of `get_latex(style=EXPRESSION)` doesn't return valid $LaTeX$, while `get_latex(style=ALGORITHMIC)` does. Feels weird to have this discrepancy, but I'm not sure of a good way to resolve it.

odashi:
General comments before providing detailed review:

- Could you split the pull request into at least 2 separate PRs: introducing IPython-specific codegen and revising frontends?
- `IPythonAlgorithmicCodegen` makes sense more than `AlgorithmicJupyterCodegen. `_repr_latex_` is the feature of IPython and Jupyter (and other REPL runtimes) borrows this interface.
- `\displaystyle` can be avoided at this point.
- `\hspace{}` with the `em` unit is better to follow the size of the characters.

odashi (commented):
Dummy, please re-request review after all changes in frontend were split into other PR.

odashi (commented):
Use `\textsc` for small capital letters.
Wrap all statements by `\begin{array}{l}` and `\end{array}`. newlines are unexpected in the bare math mode and it brings unexpected consequence (e.g., `$a\\b$` and `$$a\\b$$` may generate different results in terms of the number of lines)

ZibingZhang:
> * Use `\textsc` for small capital letters.

Tried it in the Google collab notebook and `\textsc{fact}` dosn't seem to be working

image

Maybe `\texttt`?

image

odashi:
> \textsc{fact} dosn't seem to be working

Okay, so I guess `\mathrm{...}` is eventually better, and I don't agree with any options with manual upcasing.

odashi:
> `\texttt`

No, we don't have any necessity to use typewriter font.

ZibingZhang:
> I don't agree with any options with manual uppercasing.

Sounds good, I can remove that. Only had it because `ALGORTIHMIC` codegen produces `\function{fun_name}` which auto uppercases.

src/integration_tests/integration_utils.py
```diff
@@ -26,20 +26,60 @@ def check_function(
     if not kwargs:
         latexified = frontend.function(fn)
         assert str(latexified) == latex
-        assert latexified._repr_latex_() == rf"$$ \displaystyle {latex} $$"
+        assert latexified._repr_latex_() == r"$$ \displaystyle " + latex + " $$"
```
odashi, line 29: Restore the original expression, and don't use "+" concatenation as well as possible. f-strings or `join()` are basically the most fastest option to concatenate strings and "+" is the worst choice.

Ditto for all other places.

```python
def f():
  for i in range(1_000_000):
    x = f"The answer is {i}."
  return x

def g():
  for i in range(1_000_000):
    x = "The answer is " + str(i) + "."
  return x

%time f()
%time g()
```

```
CPU times: user 191 ms, sys: 0 ns, total: 191 ms
Wall time: 192 ms
CPU times: user 327 ms, sys: 0 ns, total: 327 ms
Wall time: 327 ms
```

ZibingZhang: Good to know, I think I've converted it all back. When codegen though lots of places can't use `f` strings because of `\` not allowed :(

odashi (approved):

PR #164, opened 2022-12-11
Frontend refactor -- ipython_wrapper for decorator output, setup work for ipython-algorithmic codegen

ZibingZhang:
# Overview

Refactor frontend.

# Details

* extract `LatexifiedFunction` into `ipython_wrapper` also with a new class `LatexifiedAlgorithm` (with common base class `LatexifiedRepr`
* New style `IPYTHON_ALGORITHMIC` along with new frontend decorator `@algorithmic`

# References

The first half of this PR: https://github.com/google/latexify_py/pull/163

# Blocking

Blocking the testing of `IPythonAlgCodegen` (https://github.com/google/latexify_py/pull/163), which is why I've extracted this portion out first.

# Blocked by

None


odashi:
Title: it is somewhat hard to understand what "FE" is, and it denies to find this PR from the keyword search.

src/latexify/output.py
```diff
@@ -0,0 +1,120 @@
+"""Output of the frontend function decorators."""
+
+from __future__ import annotations
+
+import abc
+from typing import Any, Callable
+
+from latexify import exceptions, frontend
```
odashi, line 8: Circular import.

```suggestion
from latexify import exceptions
```

src/latexify/frontend.py
```diff
@@ -80,74 +95,67 @@ def get_latex(
         ).visit(tree)
 
 
-class LatexifiedFunction:
-    """Function with latex representation."""
+@overload
+def algorithmic(alg: Callable[..., Any], **kwargs: Any) -> output.LatexifiedAlgorithm:
```
odashi, line 99: Better to use the same argument name.
```suggestion
def algorithmic(fn: Callable[..., Any], **kwargs: Any) -> output.LatexifiedAlgorithm:
```

src/latexify/frontend.py
```diff
@@ -80,74 +95,67 @@ def get_latex(
...
 
-    @property
-    def __doc__(self):
-        return self._fn.__doc__
 
-    @__doc__.setter
-    def __doc__(self, val):
-        self._fn.__doc__ = val
+def algorithmic(
+    alg: Callable[..., Any] | None = None, **kwargs: Any
```
odashi, line 111:
```suggestion
    fn: Callable[..., Any] | None = None, **kwargs: Any
```

src/latexify/frontend.py
```diff
@@ -80,74 +95,67 @@ def get_latex(
...
+) -> output.LatexifiedAlgorithm | Callable[
+    [Callable[..., Any]], output.LatexifiedAlgorithm
+]:
+    """Attach LaTeX pretty-printing to the given algorithm.
```
odashi, line 115: It is unclear what kind of stuff the term "algorithm" points to. use "function".

```suggestion
    """Attach LaTeX pretty-printing to the given function.
```

src/latexify/frontend.py
```diff
@@ -80,74 +95,67 @@ def get_latex(
...
 
-    @property
-    def __name__(self):
-        return self._fn.__name__
+    This function works with or without specifying the target algorithm as the
```
odashi, line 117:
```suggestion
    This function works with or without specifying the target function as the
```

src/latexify/frontend.py
```diff
@@ -80,74 +95,67 @@ def get_latex(
...
-    def __name__(self, val):
-        self._fn.__name__ = val
+    Args:
+        alg: Callable to be wrapped.
+        **kwargs: Arguments to control behavior. See also get_latex().
 
-    def __call__(self, *args):
-        return self._fn(*args)
+    Returns:
+        - If `alg` is passed, returns the wrapped function.
```
odashi, line 127:
```suggestion
        - If `fn` is passed, returns the wrapped function.
```

src/latexify/output.py
ZibingZhang: Should I import locally in the method instead? I think we still need `frontend` because of `get_latex`.

src/latexify/transformers/transformer_utils.py
```diff
@@ -0,0 +1,30 @@
+"""Utilites for latexify.frontend."""
```
ZibingZhang: will add tests to test that appropriate transformer is called if you think this is an ok approach

src/latexify/ipython_wrappers.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+
+
+class LatexifiedFunction(LatexifiedRepr):
+    """Function with latex representation."""
+
+    _latex: str | None
+    _error: str | None
+
+    def __init__(self, fn, **kwargs):
```
ZibingZhang, line 113: bit of repeated code here and in `get_latex`, unsure of best solution to fix that

src/latexify/output.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class LatexifiedRepr(abc.ABC):
```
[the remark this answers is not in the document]
ZibingZhang: `@final` doesn't exist in python 3.7

src/latexify/ipython_wrappers.py
ZibingZhang: This definitely has to be abstracted somehow, because any new features will be duplicated in these code paths. I want to add to `Style` a value `ALGORITHMIC_IPYTHON` or something like that, then call into `get_latex` once for `Style.ALGORITHMIC` and once for `ALGORITHMIC_IPYTHON`, but that would result in a circular import. Bit stuck for solution at the moment

src/latexify/output.py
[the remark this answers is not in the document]
odashi: > `@final` doesn't exist in python 3.7

yeah, so we need to manually treat these are "final" until dropping 3.7 support

src/latexify/output.py
odashi: Circular imports can be usually resolved by separating files. In this case the problem is caused by coexisting `get_latex` and wrapper functions, so it should be resolved by splitting `get_latex` out to another file.

src/latexify/transformers/transformer_utils.py
odashi: I'm not sure what "if you think this is an ok approach" mean. what is intended here?

odashi:
Since I'm working on the company and have limited time to work on this repository, could you re-request reviews after finishing all modifications?
I could notice new comments without re-requesting review btw.

src/latexify/output.py
ZibingZhang: moved into `codegen/generate_latex.py`

ZibingZhang:
> there's no blocker in the description

Not sure what you mean by this. I don't think these changes are blocked by anything

odashi:
I meant that if this change intended to introduce only separating get_latex, it is weird that there is no blocked_by items in the description.

src/latexify/frontend.py
```diff
@@ -2,145 +2,74 @@
...
-        return (
-            r"$$ \displaystyle " + self._latex + " $$"
-            if self._latex is not None
-            else self._error
-        )
+    if fn is not None:
+        return ipython_wrappers.LatexifiedAlgorithm(fn, **kwargs)
+
+    def wrapper(a):
+        return ipython_wrappers.LatexifiedAlgorithm(a, **kwargs)
```
odashi, line 49: nit:
```suggestion
    def wrapper(f):
        return ipython_wrappers.LatexifiedAlgorithm(f, **kwargs)
```

src/latexify/generate_latex.py
```diff
@@ -0,0 +1,83 @@
...
+
+from latexify import codegen
+from latexify import config as cfg
+from latexify import parser, transformers
+
+# NOTE(odashi):
+# These prefixes are trimmed by default.
+# This behavior shouldn't be controlled by users in the current implementation because
+# some processes expects absense of these prefixes.
+_COMMON_PREFIXES = {"math", "numpy", "np"}
```
odashi, line 17: #166 is not reflected.

src/latexify/ipython_wrappers.py
```diff
@@ -0,0 +1,135 @@
...
+
+from __future__ import annotations
+
+import abc
+from typing import Any, Callable
+
+from latexify import exceptions, generate_latex
+
+
+class LatexifiedRepr(metaclass=abc.ABCMeta):
```
odashi, line 11: Add type hints for every function.

src/latexify/ipython_wrappers.py
```diff
@@ -0,0 +1,135 @@
...
+
+    @__name__.setter
+    def __name__(self, val):
+        self._fn.__name__ = val
+
+    # After Python 3.7
+    # @final
+    def __call__(self, *args):
+        return self._fn(*args)
+
```
odashi, line 39: `__str__` should also be an abstractmethod.

src/latexify/ipython_wrappers.py
```diff
@@ -0,0 +1,135 @@
...
+            self._latex = generate_latex.get_latex(
+                fn, style=generate_latex.Style.ALGORITHMIC, **kwargs
+            )
+            self._error = None
+        except exceptions.LatexifyError as e:
+            self._latex = None
+            self._error = f"{type(e).__name__}: {str(e)}"
+
+        try:
+            self._ipython_latex = self._latex = generate_latex.get_latex(
```
odashi, line 72:
```suggestion
            self._ipython_latex = generate_latex.get_latex(
```

src/latexify/generate_latex.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Generates LaTeX.
+    if style == Style.ALGORITHMIC:
+        return codegen.AlgorithmicCodegen(
+            use_math_symbols=merged_config.use_math_symbols,
+            use_set_symbols=merged_config.use_set_symbols,
+        ).visit(tree)
+    elif style == Style.IPYTHON_ALGORITHMIC:
+        # TODO(ZibingZhang): implement algorithmic codegen for ipython
+        raise NotImplementedError
```
odashi, line 75: Should raise LatexifyNotSupportedError, or LatexifiedAlgorithm always fails.

odashi (commented):
Only trivial fixes

src/latexify/ipython_wrappers.py
odashi: `_repr_something_` should return `str | tuple[str, dict[str, Any]] | None` by design, where the second case involves [metadata](https://ipython.readthedocs.io/en/stable/config/integrating.html#metadata)

src/latexify/ipython_wrappers.py
```diff
@@ -69,26 +73,33 @@ def __init__(self, fn, **kwargs):
...
-        return self._latex if self._latex is not None else self._error
+    def __str__(self) -> str:
+        if self._latex is not None:
+            return self._latex
+        elif self._error is not None:
+            return self._error
+        else:
+            raise exceptions.LatexifyError(
+                "Unable to get a string representation of the function."
+            )
```
odashi, line 92: Use `typing.cast`.
```suggestion
        return self._latex if self._latex is not None else cast(str, self._error)
```

src/latexify/ipython_wrappers.py
```diff
@@ -115,18 +126,25 @@ def __init__(self, fn, **kwargs):
...
```
odashi, line 137:
```suggestion
        return self._latex if self._latex is not None else cast(str, self._error)
```

odashi (commented):
Maybe this is the last

odashi:
Hey @ZibingZhang , I noticed that you are committing with at least 2 different accounts. In most cases this means that your local environment does not set the account information correctly. Please check how you set `git config` on your machine.

ZibingZhang:
> Hey @ZibingZhang , I noticed that you are committing with at least 2 different accounts. In most cases this means that your local environment does not set the account information correctly. Please check how you set `git config` on your machine.

Just checked and yup my email address wasn't the same as my GitHub account, thanks for letting me know! I've just fixed it, and it shouldn't be a problem for future commits / PRs.

odashi (approved):

PR #167, opened 2022-12-12
Improve algorithmic codegen, return multiline string

ZibingZhang:
# Overview

Algorithmic codegen now returns a multiline string instead of a very long single line string.

# Details

We keep track of our indentation level as we traverse through the AST.

### Example

old
```text
\If{$x < y$} \State \Return $x$ \EndIf
```

new
```text
If{$x < y$}
    \State \Return $x$
\EndIf
```

# References

https://github.com/google/latexify_py/issues/57

# Blocked by

None

### DIFF
--- a/src/latexify/codegen/algorithmic_codegen.py
+++ b/src/latexify/codegen/algorithmic_codegen.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import ast
 import contextlib
-from typing import Generator
+from collections.abc import Generator
 
 from latexify import exceptions
 from latexify.codegen import expression_codegen, identifier_converter
--- a/src/latexify/codegen/algorithmic_codegen.py
+++ b/src/latexify/codegen/algorithmic_codegen.py
@@ -68,37 +68,33 @@ class AlgorithmicCodegen(ast.NodeVisitor):
         ]
 
         latex = self._add_indent("\\begin{algorithmic}\n")
-        self._indent_level += 1
-        latex += self._add_indent(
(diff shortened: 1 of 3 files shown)


src/latexify/codegen/algorithmic_codegen.py
```diff
@@ -102,5 +120,14 @@ def visit_While(self, node: ast.While) -> str:
...
+        self._indent += 1
+        body_latex = "\n".join(self.visit(stmt) for stmt in node.body)
+        self._indent -= 1
+        return (
+            f"{self._prefix()}\\While{{${cond_latex}$}}\n"
+            f"{body_latex}\n"
+            rf"{self._prefix()}\EndWhile"
+        )
+
+    def _prefix(self) -> str:
```
odashi, line 132: - `prefix` doesn't make sense, use other name.
- Add docstring.

src/latexify/codegen/algorithmic_codegen.py
```diff
@@ -3,6 +3,8 @@
 from __future__ import annotations
 
 import ast
+import contextlib
+from typing import Generator
```
odashi, line 7: Use `collections.abc` rather than `typing` for generic types.

```suggestion
from collections.abc import Generator
```

odashi (approved):

src/latexify/codegen/algorithmic_codegen.py
ZibingZhang: just curious as to why not use typing

odashi:
LGTM, merge this PR

PR #172, opened 2022-12-15
Fix function name bug on algorithmic

odashi:
# Overview

CC: @ZibingZhang 

I noticed that there is the same bug `foo_bar` -> $foo_bar$ in generated function names by `*AlgorithmicCodegen`. This PR applies the same identifier rule to the function name as well.

Since the `mathrm` wrapper of the identifier is not necessary in `AlgorithmicCodegen`, this change also adds a flag on `IdentifierConverter` to control this behavior.

# References

NA

# Blocked by

NA


ShigekiKarita (approved):
lgtm

PR #195, opened 2023-11-18
Fix wrapping around power operators.

### LINKED ISSUE #189: Counterintuitive (wrong?) parenthesis when combining exp() and powers
## Description

When you latexify something like log((x/c)^2), the rendering you get looks (to me at least) like it's log(x/c)^2:
image

In contrast, Wolfram Alpha for example will insert extra parens to disambiguate this case.
image

## Reproduction

Describe how to reproduce the issue by other people.

```
@latexify.function()
def f(x, c):
  return np.log((x/c)**2)
f
```

odashi:
# Overview

This PR changes the wrapping behavior around function calls with pow.

# Details

Due to its visual ambiguity, the power operator should be always wrapped by brackets if it appears in the unary function.

Before:
a

After:
無題

# References

- Fix #189 

# Blocked by

### DIFF
--- a/src/latexify/codegen/expression_codegen.py
+++ b/src/latexify/codegen/expression_codegen.py
@@ -408,16 +408,22 @@ class ExpressionCodegen(ast.NodeVisitor):
 
         if rule.is_unary and len(node.args) == 1:
             # Unary function. Applies the same wrapping policy with the unary operators.
+            precedence = expression_rules.get_precedence(node)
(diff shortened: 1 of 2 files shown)


odashi:
@tesknight @xiziqiao Thanks for making very helpful conversation on #189 !
Please take a look at this change and put comments if you found something wrong.

ShigekiKarita (approved):

Issue #198, opened 2023-12-05
Include .tex output examples in latexify_py/examples/examples.ipynb 


ctarnold:
## Description

In latexify_py/examples/examples.ipynb, it would be useful to include in one of the examples the option to print to a .tex file. 

## Ideas of the solution

This feature could involve redefining stdout to a .tex file, or by opening a new .tex file and writing the output of a latexify function to the new .tex file.

## Additional context

This feature would allow more ease of use for the client who wishes to paste the tex output to a Latex editor.

odashi:
Thanks @ctarnold ! Could you explain more details you are thinking of?

I think the obvious way is to print the wrapped function directly as follows because `str()` of the wrapped function prints the underlying source.

```python
@latexify.function
def f():
  return 42

with open("some.tex", "w") as fp:
    print(f, file=fp)
```

The output would contain:

```tex
f() = 42
```

However, this is not beyond the usual practice in Python and I'm wondering whether it is really necessary to note.

ctarnold:
Hi @odashi, thank you for your response! I came to this repository as a potential user - I love the product. I initially envisioned the use case of exporting python functions to a research paper drafted in latex. I exist as a subset of users who are much more familiar with latex than they are with python. 

I'm considering the example use-case of someone, perhaps in academia, who wants to export a colleague's code to a research draft.  That said, your proposed example for exporting to a file would be a better fix. I also tend to agree that for most users familiar with python this example would be trivial. I think this change is minor, but also better communicates this product to others. Let me know your thoughts, and I'm open to modifying my pull-request or resubmitting.

PR #201, opened 2023-12-07
Renew example notebook

odashi:
# Overview

# Details

- Renewal of the whole content in `examples/examples.ipynb` with renaming it to `examples/latexify_examples.ipynb`.
- The change also includes #198
- Replace the Colab link so that the above file is loaded rather than another notebook managed on Colab.

# References

# Blocked by


ShigekiKarita (approved):
Nice examples!

PR #208, opened 2024-05-13
Escape underscores

bjude:
# Overview

Add the option to not escape underscores in IdentifierConverter

# Details

This allows for generating symbols with subscripts (possibly multiple layers), e.g. x_2_i becomes x_{2_{i}}. This is done by splitting the identifier on underscores, converting each 'sub-identifier' as normal, then concatenating the converted identifiers back together, separated by underscores. There is some simple logic in the concatenation to wrap the subscripts in braces to accommodate multiple layers of subscripts.

This option also creates the possibility of generating invalid latex if the identifier starts or ends with an underscore, or has a double underscore in it. In these cases we now raise a ValueError. Tests have been added exercising the new functionality, including the error path.

### DIFF
--- a/src/latexify/codegen/identifier_converter.py
+++ b/src/latexify/codegen/identifier_converter.py
@@ -58,20 +58,11 @@ class IdentifierConverter:
         if not self._escape_underscores and "_" in name:
             # Check if we are going to generate an invalid Latex string. Better to raise an
             # exception here than have the resulting Latex fail to compile/display
-            if "__" in name:
+            name_splits = name.split("_")
+            if not all(name_splits):
                 raise ValueError(
-                    f'Identifier "{name}" has a double underscore will result in '
-                    "invalid Latex when underscores are not escaped"
-                )
-            if name.startswith("_"):
-                raise ValueError(
-                    f'Identifier "{name}" starts with an underscore will result in '
-                    "invalid Latex when underscores are not escaped"
-                )
-            if name.endswith("_"):
-                raise ValueError(
-                    f'Identifier "{name}" ends with an underscore will result in '
-                    "invalid Latex when underscores are not escaped"
+                    "Neither preceding/trailing underscores nor double underscores is allowed "
+                    f"by the `escape_underscores` option, but got: {name}"
                 )
             elems = [
                 IdentifierConverter(
@@ -79,7 +70,7 @@ class IdentifierConverter:
                     use_mathrm=False,
                     escape_underscores=True,
                 ).convert(n)[0]
-                for n in name.split("_")
+                for n in name_splits
             ]
             # Wrap sub identifiers in nested braces
             name = "_{".join(elems) + "}" * (len(elems) - 1)
--- a/src/latexify/codegen/identifier_converter.py
+++ b/src/latexify/codegen/identifier_converter.py
@@ -33,8 +33,8 @@ class IdentifierConverter:
                 appropriate LaTeX command.
             use_mathrm: Whether to wrap the resulting expression by \mathrm, if
                 applicable.
-            escape_underscores: Whether to prefix any underscores in identifiers with '\\',
-                disable to allow subscripts in generated latex
+            escape_underscores: Whether to prefix any underscores in identifiers with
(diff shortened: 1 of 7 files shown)


subsurfaceiodev:
This feature would be awesome

src/latexify/codegen/identifier_converter.py
```diff
@@ -41,14 +51,51 @@ def convert(self, name: str) -> tuple[str, bool]:
...
+            if name.startswith("_"):
+                raise ValueError(
+                    f'Identifier "{name}" starts with an underscore will result in '
+                    "invalid Latex when underscores are not escaped"
+                )
+            if name.endswith("_"):
+                raise ValueError(
+                    f'Identifier "{name}" ends with an underscore will result in '
+                    "invalid Latex when underscores are not escaped"
+                )
```
odashi: I thought that these all checks can be performed by just checking if the split contains empty fragment:
```python
name_splits = name.split("_")
if not all(name_splits):
    # some fragment is empty
    raise ValueError(...)
elms = [... for n in name_splits]
```

src/latexify/codegen/identifier_converter.py
```diff
@@ -41,14 +51,51 @@ def convert(self, name: str) -> tuple[str, bool]:
...
+            elems = [
+                IdentifierConverter(
+                    use_math_symbols=self._use_math_symbols,
+                    use_mathrm=False,
+                    escape_underscores=True,
+                ).convert(n)[0]
```
odashi, line 73:
```suggestion
                self.convert(n)[0]
```

src/latexify/codegen/identifier_converter.py
```diff
@@ -41,14 +51,51 @@ def convert(self, name: str) -> tuple[str, bool]:
...
+                for n in name.split("_")
+            ]
+            if name.count("_") == 1:
+                # One underscore, no need to wrap subsequent subscripts in braces
+                name = "_".join(elems)
+            else:
+                # Multi-layer subscripts require wrapping each subscript in nesting
+                # braces
+                name = "_{".join(elems) + "}" * (len(elems) - 1)
```
odashi: The else-case actually covers all number of elements, and we need to wrap the 2nd elm even when `len(elms) == 2` (e.g., `a_bc` should be `a_{bc}`)
```suggestion
            name = "_{".join(elems) + "}" * (len(elems) - 1)
```

src/latexify/codegen/identifier_converter.py
bjude: It could, we would still need a series of `if` statements to give good error messages in the exception(s) so no code would be saved

src/latexify/codegen/identifier_converter.py
bjude: Can't use `self` here as we need to ensure `use_mathrm` is `False` when processing the sub-identifiers, we only want mathrm{} to be wrapped around the whole identifier, not each sub-identifier. `escape_underscores=True` was added to prevent an infinite recursion and to ensure line 98 triggers correctly

src/latexify/codegen/identifier_converter.py
bjude: Correct, the tests i wrote were wrong when rendered as latex.
Accepted and tests updated

src/latexify/codegen/identifier_converter.py
odashi: ah okay my bad. thanks for clarifying.

src/latexify/codegen/identifier_converter.py
odashi: I think it would be possible to cover all illegal cases in a single message. How about:

```python
Neither preceding/trailing underscores nor double underscores is allowed by the `escape_underscores` option, but got: {name}
```

bjude: Yeah that works, the identifiers should be clearly visible in the editor so a specific error message probably isnt needed.

odashi (approved):
Thanks!

odashi:
@bjude I would merge this PR after all CIs passed.

PR #210, opened 2024-10-08
Pump Version python to 3.12, 3.13. Allow pip install for python 3.12, 3.13

maycuatroi:
# Overview

Update file `pyproject.toml` to enable pip install for python 3.12

# References

Example workbook work perfect when I pip install on python 3.12
image

### DIFF
--- a/src/latexify/ast_utils.py
+++ b/src/latexify/ast_utils.py
@@ -57,16 +57,10 @@ def make_constant(value: Any) -> ast.expr:
         ValueError: Unsupported value type.
     """
     if sys.version_info.minor < 8:
-        if value is None or value is False or value is True:
-            return ast.NameConstant(value=value)
-        if value is ...:
-            return ast.Ellipsis()
-        if isinstance(value, (int, float, complex)):
-            return ast.Num(n=value)
-        if isinstance(value, str):
-            return ast.Str(s=value)
-        if isinstance(value, bytes):
-            return ast.Bytes(s=value)
+        if value in {None, False, True, ...} or isinstance(
+            value, (int, float, complex, str, bytes)
+        ):
+            return ast.Constant(value=value)
     else:
         if (
             value is None
--- a/src/latexify/ast_utils.py
+++ b/src/latexify/ast_utils.py
@@ -147,3 +147,9 @@ def extract_function_name_or_none(node: ast.Call) -> str | None:
         return node.func.attr
 
     return None
+
+
+def ast_function_def(*args, **kwargs) -> ast.FunctionDef:
+    if sys.version_info.minor < 12:
+        kwargs.pop("type_params", None)
+    return ast.FunctionDef(*args, **kwargs)
--- a/src/latexify/ast_utils.py
+++ b/src/latexify/ast_utils.py
@@ -149,7 +149,19 @@ def extract_function_name_or_none(node: ast.Call) -> str | None:
     return None
 
 
-def ast_function_def(*args, **kwargs) -> ast.FunctionDef:
+def create_function_def(*args, **kwargs) -> ast.FunctionDef:
+    """Creates a FunctionDef node.
+
+    This function generates an `ast.FunctionDef` node, optionally removing
+    the `type_params` keyword argument for Python versions below 3.12.
+
+    Args:
+        *args: Positional arguments for `ast.FunctionDef`.
+        **kwargs: Keyword arguments for `ast.FunctionDef`.
+
+    Returns:
+        ast.FunctionDef: The generated FunctionDef node.
+    """
     if sys.version_info.minor < 12:
         kwargs.pop("type_params", None)
     return ast.FunctionDef(*args, **kwargs)
--- a/src/latexify/ast_utils.py
+++ b/src/latexify/ast_utils.py
@@ -149,19 +149,63 @@ def extract_function_name_or_none(node: ast.Call) -> str | None:
     return None
 
 
-def create_function_def(*args, **kwargs) -> ast.FunctionDef:
+def create_function_def(
+    name,
+    args,
+    body,
+    decorator_list,
+    returns=None,
+    type_comment=None,
+    type_params=None,
+    lineno=None,
+    col_offset=None,
+    end_lineno=None,
+    end_col_offset=None,
+) -> ast.FunctionDef:
     """Creates a FunctionDef node.
 
     This function generates an `ast.FunctionDef` node, optionally removing
     the `type_params` keyword argument for Python versions below 3.12.
 
     Args:
-        *args: Positional arguments for `ast.FunctionDef`.
-        **kwargs: Keyword arguments for `ast.FunctionDef`.
+        name: Name of the function.
+        args: Arguments of the function.
+        body: Body of the function.
+        decorator_list: List of decorators.
+        returns: Return type of the function.
+        type_comment: Type comment of the function.
+        type_params: Type parameters of the function.
+        lineno: Line number of the function definition.
+        col_offset: Column offset of the function definition.
+        end_lineno: End line number of the function definition.
+        end_col_offset: End column offset of the function definition.
 
     Returns:
         ast.FunctionDef: The generated FunctionDef node.
     """
     if sys.version_info.minor < 12:
-        kwargs.pop("type_params", None)
-    return ast.FunctionDef(*args, **kwargs)
+        return ast.FunctionDef(
+            name=name,
+            args=args,
+            body=body,
+            decorator_list=decorator_list,
+            returns=returns,
+            type_comment=type_comment,
+            lineno=lineno,
+            col_offset=col_offset,
+            end_lineno=end_lineno,
+            end_col_offset=end_col_offset,
+        )  # type: ignore
+    return ast.FunctionDef(
+        name=name,
+        args=args,
+        body=body,
+        decorator_list=decorator_list,
+        returns=returns,
+        type_comment=type_comment,
+        type_params=type_params,
+        lineno=lineno,
+        col_offset=col_offset,
+        end_lineno=end_lineno,
+        end_col_offset=end_col_offset,
+    )  # type: ignore
--- a/src/latexify/ast_utils.py
+++ b/src/latexify/ast_utils.py
@@ -56,18 +56,12 @@ def make_constant(value: Any) -> ast.expr:
     Raises:
         ValueError: Unsupported value type.
     """
-    if sys.version_info.minor < 8:
-        if value in {None, False, True, ...} or isinstance(
-            value, (int, float, complex, str, bytes)
-        ):
-            return ast.Constant(value=value)
-    else:
-        if (
-            value is None
-            or value is ...
-            or isinstance(value, (bool, int, float, complex, str, bytes))
-        ):
-            return ast.Constant(value=value)
+    if (
+        value is None
+        or value is ...
+        or isinstance(value, (bool, int, float, complex, str, bytes))
+    ):
+        return ast.Constant(value=value)
 
     raise ValueError(f"Unsupported type to generate Constant: {type(value).__name__}")
 
@@ -81,13 +75,7 @@ def is_constant(node: ast.AST) -> bool:
     Returns:
         True if the node is a constant, False otherwise.
     """
-    if sys.version_info.minor < 8:
-        return isinstance(
-            node,
-            (ast.Bytes, ast.Constant, ast.Ellipsis, ast.NameConstant, ast.Num, ast.Str),
-        )
-    else:
-        return isinstance(node, ast.Constant)
+    return isinstance(node, ast.Constant)
 
 
 def is_str(node: ast.AST) -> bool:
@@ -114,20 +102,12 @@ def extract_int_or_none(node: ast.expr) -> int | None:
     Returns:
         Extracted int value, or None if extraction failed.
     """
-    if sys.version_info.minor < 8:
-        if (
-            isinstance(node, ast.Num)
-            and isinstance(node.n, int)
-            and not isinstance(node.n, bool)
-        ):
-            return node.n
-    else:
-        if (
-            isinstance(node, ast.Constant)
-            and isinstance(node.value, int)
-            and not isinstance(node.n, bool)
-        ):
-            return node.value
+    if (
+        isinstance(node, ast.Constant)
+        and isinstance(node.value, int)
+        and not isinstance(node.value, bool)
+    ):
+        return node.value
 
     return None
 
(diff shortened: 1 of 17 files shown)


odashi (approved):

maycuatroi:
It's strange, I didn't make any changes to the Python code, yet it's causing CI errors with mypy. 
@odashi-san, Do you think I should fix these CI errors?

maycuatroi:
@odashi-san. I fixed CI to the last version of mypy and black. Please approve the CI scan

odashi:
@maycuatroi Thanks!

I think you tweaked some code blocked by version switches, e.g.,

* if blocks conditioned by `if sys.version_info.minor < blah:`
* test blocks conditioned by `@test_utils.require_at_most(blah)`

These blocks are activated while the library is running on a conditioned version of Python, and the code block must follow the correct syntax on the corresponding version.

I guessed that this is a good time to drop support for legacy (EOL-ed) versions such as 3.7 and 3.8, by simply removing corresponding blocks from the library.
Sorry for bothering, but I'd quite appreciated it if you could take some additional efforts.

maycuatroi:
@odashi -san
I've checked and fixed the unit tests for the CI accordingly. Here's what I did:

- Removed support for Python 3.7 and 3.8 (since they've reached end-of-life).
- Tested with Python 3.13 – everything worked great!

I really enjoy using this library, and I appreciate your professionalism.

To ensure everything works across multiple Python versions, I also created a custom Docker Compose setup for running the CI tests on various versions 😄  (I'm not commit it yet):

```yaml
version: "3.8"

services:
  unit-tests:
    image: python:3.13
    environment:
      - PYTHONPATH=src
    volumes:
      - .:/app  # This will copy all files in the current directory into the /app directory in the container
    working_dir: /app
    command: >
      bash -c "
      python -m pip install --upgrade pip &&
      python -m pip install -e '.[dev]' &&
      python -m pytest src
      "

  black:
    image: python:3.13
    volumes:
      - .:/app  # Mounting the project directory
    working_dir: /app
    command: >
      bash -c "
      python -m pip install --upgrade pip &&
      python -m pip install black &&
      python -m black -v --check src
      "

  flake8:
    image: python:3.13
    volumes:
      - .:/app  # Mounting the project directory
    working_dir: /app
    command: >
      bash -c "
      python -m pip install --upgrade pip &&
      python -m pip install pyproject-flake8 &&
      pflake8 -v src
      "

  isort:
    image: python:3.13
    volumes:
      - .:/app  # Mounting the project directory
    working_dir: /app
    command: >
      bash -c "
      python -m pip install --upgrade pip &&
      python -m pip install isort &&
      python -m isort --check src
      "

  mypy:
    image: python:3.13
    volumes:
      - .:/app  # Mounting the project directory
    working_dir: /app
    command: >
      bash -c "
      python -m pip install --upgrade pip &&
      python -m pip install '.[mypy]' &&
      python -m mypy src
      "
```

src/latexify/test_utils.py
```diff
@@ -93,8 +97,8 @@ def ast_equal(observed: ast.AST, expected: ast.AST) -> bool:
                 assert type(vo) is type(ve)
                 assert vo == ve
 
-    except (AssertionError, AttributeError):
-        return False
+    except (AssertionError, AttributeError) as e:
+        raise e  # raise to debug easier.
```
odashi, line 101:
```suggestion
    except (AssertionError, AttributeError):
        raise  # raise to debug easier.
```

src/latexify/test_utils.py
```diff
@@ -109,19 +113,7 @@ def assert_ast_equal(observed: ast.AST, expected: ast.AST) -> None:
...
-            observed, expected
-        ), f"""\
-AST does not match.
-observed={ast.dump(observed)}
-expected={ast.dump(expected)}
-"""
+    indent = 4 if sys.version_info.minor >= 9 else None
+    assert ast_equal(
+        observed, expected
+    ), f"AST does not match. observed = {observed}\nexpected = {expected}"
```
odashi, line 119: I looks this change is incomplete and should be reverted for now

src/latexify/ast_utils.py
```diff
@@ -57,16 +57,10 @@ def make_constant(value: Any) -> ast.expr:
...
-            return ast.Num(n=value)
-        if isinstance(value, str):
-            return ast.Str(s=value)
-        if isinstance(value, bytes):
-            return ast.Bytes(s=value)
+        if value in {None, False, True, ...} or isinstance(
+            value, (int, float, complex, str, bytes)
+        ):
+            return ast.Constant(value=value)
     else:
```
odashi, line 64: As we decided to drop old versions, this if-block is no longer necessary.

src/latexify/ast_utils_test.py
```diff
@@ -38,15 +38,15 @@ def test_make_attribute() -> None:
 @pytest.mark.parametrize(
```
odashi, line 38: Ditto: as this test case only runs on older versions than 3.8, we can simply remove it.

src/latexify/transformers/identifier_replacer.py
```diff
@@ -55,21 +55,25 @@ def visit_FunctionDef(self, node: ast.FunctionDef) -> ast.FunctionDef:
                 kwonlyargs=self._replace_args(visited.args.kwonlyargs),
                 kw_defaults=visited.args.kw_defaults,
                 defaults=visited.args.defaults,
-            )
+            )  # type: ignore
```
odashi, line 58: Could you resolve typing errors?
If it is difficult, could you use `#type: ignore[code]` rather than the wildcard `#type: ignore` to show which error code we ignored?

maycuatroi: I removed the python 3.8 conditions

src/latexify/test_utils.py
maycuatroi: I removed the if condition because python always >=3.9 and revert the ast dump logic

src/latexify/ast_utils_test.py
maycuatroi: I’m not completely sure I understand this part. Could you help me clarify or commit to update? 
I noticed there were some test cases added earlier, but I’m unsure of their purpose, so I didn’t want to remove them just yet. I’ve updated the code to stop using the deprecated functions based on warnings from the latest version, but I wanted to double-check with you first.

maycuatroi:
@odashi  -san
I Create new function `ast_function_def` to handle `type_params` for python 3.12 and I fixed the flake8 CI issue. Please help me review and approve workflow 🙏

src/latexify/ast_utils_test.py
maycuatroi: I removed this test case

maycuatroi:
@odashi -san. I think things look good 😄

src/latexify/ast_utils.py
```diff
@@ -147,3 +147,9 @@ def extract_function_name_or_none(node: ast.Call) -> str | None:
         return node.func.attr
 
     return None
+
+
+def ast_function_def(*args, **kwargs) -> ast.FunctionDef:
```
odashi, line 152: Please follow the convention of this library:

* function names should be basically starting with a verb
* add docstring (with Google's docstring style)
* add unit test for this function (in `ast_utils_test.py`).
* (if possible) full argument list is encouraged rather than `args` and `kwargs`.

src/latexify/ast_utils.py
```diff
@@ -147,3 +147,9 @@ def extract_function_name_or_none(node: ast.Call) -> str | None:
...
+    if sys.version_info.minor < 12:
+        kwargs.pop("type_params", None)
+    return ast.FunctionDef(*args, **kwargs)
```
odashi, line 155: I think we haven't use `type_params` anywhere with meaningful semantics, and this function can be defined to hide this variable from outer functions rather than passing it always.

src/latexify/ast_utils_test.py
```diff
@@ -83,7 +81,6 @@ def test_make_constant_invalid() -> None:
         ast_utils.make_constant(object())
 
 
-@test_utils.require_at_most(7)
 @pytest.mark.parametrize(
```
odashi, line 84: ditto: this test case can be removed.

src/latexify/ast_utils_test.py
```diff
@@ -114,7 +110,6 @@ def test_is_constant(value: ast.AST, expected: bool) -> None:
     assert ast_utils.is_constant(value) is expected
 
 
-@test_utils.require_at_most(7)
 @pytest.mark.parametrize(
```
odashi, line 113: ditto

src/latexify/ast_utils_test.py
```diff
@@ -34,7 +34,6 @@ def test_make_attribute() -> None:
     )
 
 
-@test_utils.require_at_most(7)
```
odashi, line 37: This test case (the whole function) is no longer meaningful, and should be completely removed.

Same as every other test case with a decoration `@test_utils.require_at_most(7)`

src/latexify/parser_test.py
```diff
@@ -7,6 +7,7 @@
 import pytest
 
 from latexify import exceptions, parser, test_utils
+from latexify.ast_utils import ast_function_def
```
odashi, line 10: We don't allow importing functions/classes directly, import module instead.

Same for every other code.

Reference: https://google.github.io/styleguide/pyguide.html#22-imports

```suggestion
from latexify import ast_utils
```

odashi (commented):
Cool, basically everything looks good. Comments are almost based on coding conventions.

src/latexify/ast_utils_test.py
maycuatroi: I removed the legacy test case

src/latexify/ast_utils.py
maycuatroi: We need `type_params` to pass the `mypy` scan

src/latexify/ast_utils.py
maycuatroi: Yes, perfect. I updated the coding convention and created the test case.
Kwargs current can't replace be cause `type_params` is not optional so we need pass through kwargs and filter in the wrapper function

src/latexify/ast_utils.py
odashi: How about:

```python
def this_fn(...):
    if sys.version_info.minor < 12:
        # We know that the following invocation is correct.
        # In this case the mypy error is just wrong and we need to ignore it.
        return ast.FunctionDef(...)  # type: ignore[some-appropriate-flag]
  
    return ast.FunctionDef(..., type_params=[])
```

maycuatroi: I think mark ignore on this function is better. Let me update

maycuatroi:
@odashi  -san, please approve the workflow, I fixed the isort issue 😄

src/latexify/ast_utils.py
maycuatroi: I updated follow your comment

maycuatroi:
@odashi -san, please help me review. I updated follow your comment 😃

src/latexify/ast_utils_test.py
```diff
@@ -83,25 +61,23 @@ def test_make_constant_invalid() -> None:
...
+        (ast.Constant(value="bar"), True),
+        (ast.Constant(value=...), True),
+        (ast.Constant(value=None), True),
+        (ast.Constant(value=123), True),
+        (ast.Constant(value="baz"), True),
+        (ast.Expr(value=ast.Constant(value=456)), False),
         (ast.Global(names=["qux"]), False),
     ],
 )
 def test_is_constant_legacy(value: ast.AST, expected: bool) -> None:
```
odashi, line 77: It looks there exists some test cases from that only `@test_utils.require_at_most(7)` was removed like this. Could you check the changes entirely and remove them?

src/latexify/ast_utils.py
```diff
@@ -173,3 +147,65 @@ def extract_function_name_or_none(node: ast.Call) -> str | None:
...
+    args,
+    body,
+    decorator_list,
+    returns=None,
+    type_comment=None,
+    type_params=None,
+    lineno=None,
+    col_offset=None,
+    end_lineno=None,
+    end_col_offset=None,
```
odashi, line 163: please add type hints.

src/latexify/ast_utils_test.py
```diff
@@ -114,25 +90,23 @@ def test_is_constant(value: ast.AST, expected: bool) -> None:
...
+        (ast.Constant(value="bar"), True),
+        (ast.Constant(value=...), False),
+        (ast.Constant(value=None), False),
+        (ast.Constant(value=123), False),
+        (ast.Constant(value="baz"), True),
+        (ast.Expr(value=ast.Constant(value=456)), False),
         (ast.Global(names=["qux"]), False),
     ],
 )
 def test_is_str_legacy(value: ast.AST, expected: bool) -> None:
```
odashi, line 106: this test case as well

odashi (commented):
It's almost done I think, but there are still some little comments (maybe the last ones)

src/latexify/ast_utils_test.py
maycuatroi: The decorator (@test_utils.require_at_most(7)) was previously used to indicate that the test case would run on Python versions >= 3.7. So, even when we were on Python 3.9, the test case was still executed. Now that we've upgraded to at least Python 3.9, the test case should still run as intended. I believe it’s best to keep the test case as it is.

src/latexify/ast_utils_test.py
maycuatroi: Same with decorator @test_utils.require_at_most(7)

src/latexify/ast_utils_test.py
odashi: `require_at_most(k)` requires versions equal or *less* than 3.k. the decorator does nothing otherwise.
https://github.com/google/latexify_py/blob/678cd0e463ee0cfd179fa0da2897e7e0535c3b46/src/latexify/test_utils.py#L51-L52

src/latexify/ast_utils_test.py
odashi: Yes, removing every test function with decoration `@test_utils.require_at_most(7)` should be fine.

maycuatroi: @odashi  Yes, I removed all the test function decorated with @test_utils.require_at_most(7)

maycuatroi: @odashi  please check and approve the Checks :D

odashi (approved):
LGTM

PR #216, opened 2026-03-02
Allow Python keywords as identifier replacement values

### LINKED ISSUE #214: Support greek lambda character
First off, thanks for this great package! I'm not sure if this would qualify as a bug or a feature, but it would be great to get access to the $\lambda$ character using `latexify`. 

## Environment

* OS: Ubuntu 24.04
* Python: 3.12
* Package manager: pip 24.2.1
* Latexify version: latest (https://github.com/google/latexify_py/commit/54dc8697196370dad227b0d516b42f6b82825852)

## Description

I would like to use the $\lambda$ symbol in my formulas but currently that raises an error. 

## Reproduction

```python
import latexify

@latexify.function(identifiers={"foo": "f", "lambda_": "lambda"}, use_math_symbols=True)
def foo(lambda_):
    return lambda_ + 1

foo

This gives me a `ValueError: 'lambda' is not an identifier name.`.
```

## Expected behavior

I would expect to be allowed to use the lambda symbol in LaTeX.

If I comment out these two lines, e.g. it does not throw an error when lambda is used. Obviously those line are probably there for a reason, so I'm not sure what the consequences are when they are removed, but perhaps the restrictions on the value side of the identifiers could be a bit more lenient...?

https://github.com/google/latexify_py/blob/54dc8697196370dad227b0d516b42f6b82825852/src/latexify/transformers/identifier_replacer.py#L42-L43

Screenshot of the output when those lines are commented out.

Image

eyupcanakman:
# Overview

Remove the `keyword.iskeyword()` check on replacement values in `IdentifierReplacer` so that keywords like `"lambda"` can be used as identifier mappings.

# Details

The `identifiers` parameter lets users map Python variable names to LaTeX symbol names (e.g., `{"lambda_": "lambda"}`). Combined with `use_math_symbols=True`, `"lambda"` renders as `\lambda`.

The validation in `IdentifierReplacer.__init__` rejected Python keywords for both keys and values. The key check is correct since keys reference real AST identifiers, but the value check was too strict. Values are replacement strings that get consumed by the codegen as plain strings for LaTeX output, so there is no reason to block keywords.

I removed the `keyword.iskeyword(v)` condition from the value validation, updated the docstring to reflect the new contract, and added tests covering keyword value acceptance at both the unit and integration level.

# References

Fixes #214

# Blocked by

N/A


eyupcanakman:
@google-cla check again

